PR #8, opened 2015-11-12
Add logger plugin

joelwurtz:
Actually don't know if this "normalizer" should be in the plugins library, WDYT ?

### DIFF
--- a/spec/LoggerPluginSpec.php
+++ b/spec/LoggerPluginSpec.php
@@ -15,9 +15,9 @@ use Psr\Log\LoggerInterface;
 
 class LoggerPluginSpec extends ObjectBehavior
 {
-    function let(LoggerInterface $logger, Normalizer $normalizer)
+    function let(LoggerInterface $logger)
     {
-        $this->beConstructedWith($logger, $normalizer);
+        $this->beConstructedWith($logger);
     }
 
     function it_is_initializable()
@@ -30,13 +30,18 @@ class LoggerPluginSpec extends ObjectBehavior
         $this->shouldImplement('Http\Client\Plugin\Plugin');
     }
 
-    function it_logs(LoggerInterface $logger, Normalizer $normalizer, RequestInterface $request, ResponseInterface $response)
+    function it_logs_request_and_response(LoggerInterface $logger, RequestInterface $request, ResponseInterface $response)
     {
         $logger->info('Emit request: "GET / 1.1"', ['request' => $request])->shouldBeCalled();
         $logger->info('Receive response: "200 Ok 1.1" for request: "GET / 1.1"', ['request' => $request, 'response' => $response])->shouldBeCalled();
 
-        $normalizer->normalizeRequestToString($request)->willReturn('GET / 1.1');
-        $normalizer->normalizeResponseToString($response)->willReturn('200 Ok 1.1');
+        $request->getMethod()->willReturn('GET');
+        $request->getRequestTarget()->willReturn('/');
+        $request->getProtocolVersion()->willReturn('1.1');
+
+        $response->getReasonPhrase()->willReturn('Ok');
+        $response->getProtocolVersion()->willReturn('1.1');
+        $response->getStatusCode()->willReturn('200');
 
         $next = function () use ($response) {
             return new FulfilledPromise($response->getWrappedObject());
@@ -45,14 +50,16 @@ class LoggerPluginSpec extends ObjectBehavior
         $this->handleRequest($request, $next, function () {});
     }
 
-    function it_logs_exception(LoggerInterface $logger, Normalizer $normalizer, RequestInterface $request)
+    function it_logs_exception(LoggerInterface $logger, RequestInterface $request)
     {
         $exception = new NetworkException('Cannot connect', $request->getWrappedObject());
 
         $logger->info('Emit request: "GET / 1.1"', ['request' => $request])->shouldBeCalled();
         $logger->error('Error: "Cannot connect" when emitting request: "GET / 1.1"', ['request' => $request, 'exception' => $exception])->shouldBeCalled();
 
-        $normalizer->normalizeRequestToString($request)->willReturn('GET / 1.1');
+        $request->getMethod()->willReturn('GET');
+        $request->getRequestTarget()->willReturn('/');
+        $request->getProtocolVersion()->willReturn('1.1');
 
         $next = function () use ($exception) {
             return new RejectedPromise($exception);
@@ -61,7 +68,7 @@ class LoggerPluginSpec extends ObjectBehavior
         $this->handleRequest($request, $next, function () {});
     }
 
-    function it_logs_response_within_exception(LoggerInterface $logger, Normalizer $normalizer, RequestInterface $request, ResponseInterface $response)
+    function it_logs_response_within_exception(LoggerInterface $logger, RequestInterface $request, ResponseInterface $response)
     {
         $exception = new HttpException('Forbidden', $request->getWrappedObject(), $response->getWrappedObject());
 
@@ -72,8 +79,13 @@ class LoggerPluginSpec extends ObjectBehavior
             'exception' => $exception
         ])->shouldBeCalled();
 
-        $normalizer->normalizeRequestToString($request)->willReturn('GET / 1.1');
-        $normalizer->normalizeResponseToString($response)->willReturn('403 Forbidden 1.1');
+        $request->getMethod()->willReturn('GET');
+        $request->getRequestTarget()->willReturn('/');
+        $request->getProtocolVersion()->willReturn('1.1');
+
+        $response->getReasonPhrase()->willReturn('Forbidden');
+        $response->getProtocolVersion()->willReturn('1.1');
+        $response->getStatusCode()->willReturn('403');
 
         $next = function () use ($exception) {
             return new RejectedPromise($exception);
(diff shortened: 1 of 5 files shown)


src/LoggerPlugin.php
```diff
@@ -0,0 +1,74 @@
...
+                $this->logger->error(
+                    sprintf('Error: "%s" when emitting request: "%s"', $exception->getMessage(), $this->normalizer->normalizeRequestToString($request)),
+                    [
+                        'request' => $this->normalizer->normalizeRequestToArray($request),
+                        'exception' => $exception
+                    ]
+                );
+            }
+
+            throw $exception;
```
dbu, line 78: so the promise is handling the throwing?

dbu, line 78: by catching the exception, i mean.

dbu:
i think it would make sense to have an own thing for that. did you know https://github.com/namshi/cuzzle ? something like this for PSR-7 would be really cool. i would create a RequestFormatter and ResponseFormatter interface in the utils and implement it with a StringFormatter for now. and later we can do the curl plugin. the formatters would have just one method `format`, keeping things interchangeable. normalizeToString and the like are tighter coupling.
the array conversion is a special case, we might want that simply as a method in the logger plugin even.

sagikazarmark:
Do we really need this? I mean, the nornalizer could work (probably in the utils repo), but in case of Monolog for example, we could have a Monolog Formatter.

The point is that it is not really the plugin's responsibility IMHO.

Haven't know cuzzle before, but it's great.

dbu:
yeah, i really love the idea of formatting a log so you can simply paste it on the cli to see what that request returned.

dbu:
yeah, formatting belongs into utils. except the monolog (or maybe psr log?) normalizing, because that is specific to the task of logging. i think a formatter should return only string, not mixed or string|array, it would make little sense. we can also have an array normalizer in utils if we think there are other use cases, but i would not implement the Formatter interface with that.

sagikazarmark:
Actually we could even format it to plain HTTP, so that we can directly send it with local testing clients.

I think request/response formatter interfaces should go into a separate package. With some basic implementations. Also, formatters should always format to string I think.

sagikazarmark:
The logger plugin should probably not require the formatter to be injected. The underlying PSR3 implementation should provide a way to add custom formatters. Not sure if it is possible though.

dbu:
> The logger plugin should probably not require the formatter to be
> injected. The underlying PSR3 implementation should provide a way to add
> custom formatters. Not sure if it is possible though.

for the message, we need a formatter i think. but for the context i
agree that we should try to push that to the logger and just pass the
object, as we do with the $exception.

> I think request/response formatter interfaces should go into a separate
> package. With some basic implementations. Also, formatters should always
> format to string I think.

not sure if we need a separate package for that. imho utils is a good
place for this. its a psr 7 utility to format requests and responses as
strings.

sagikazarmark:
> its a psr 7 utility to format requests and responses as strings.

I have a feeling that this utils repo will eventually get a bit fat. Also, I imagine the utils repo as a collection of utilities for httplug and clients. Putting too much stuff in it is not a good idea IMO. We could have a psr7-utils repo for strictly PSR-7 utilities, without Httplug and other client related stuff in it.

dbu:
i am more afraid of too many repositories than of that thing becoming
too fat. but i can see your reasoning. having several X-utils
repositories sounds confusing too. lets make it a psr7-formatter
repository then. that way its clear what its about.

oh, or does something like that already exist?

sagikazarmark:
> having several X-utils repositories sounds confusing too.

psr7-utils, httplug-utils....I can't imagine more utils for now.

> i am more afraid of too many repositories than of that thing becoming too fat

Thinking about this: Maybe a psr7-utils package would be good, because there can be several cases when we just want one small psr7 thing to be somewhere. So a common psr7-utils repo could mitigate this problem.

> oh, or does something like that already exist?

Don't think so, at least I don't know about it.

src/LoggerPlugin.php
joelwurtz, line 78: Yes when calling one of the callable passed to the then method, if it returns something then the value will go into the next onFulfilled callable, if it throw an exception it will go into the next onRejected callable.

dbu:
fine for me.

for the question whether there are existing tools, i asked the internets
https://twitter.com/dbu/status/665090150339321856

joelwurtz:
The Formatter thing in ivory http adapter was also used for that : https://github.com/egeloen/ivory-http-adapter/blob/master/src/Event/Formatter/Formatter.php

joelwurtz:
So to make it short on the changes : 
- Remove object to array transform and use the object directly in context
- Keep string normalization for the moment (but make an issue to have something more deep, like cuzzle in the future)

Ok for everyone ? @dbu @sagikazarmark

dbu:
agreed. but the Normalizer should have an `@internal` annotation and clearly state its not meant to be reused. i guess that is still better than moving those methods into the plugin class.

sagikazarmark:
:+1:

spec/LoggerPluginSpec.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        $this->shouldHaveType('Http\Client\Plugin\LoggerPlugin');
+    }
+
+    function it_is_a_plugin()
+    {
+        $this->shouldImplement('Http\Client\Plugin\Plugin');
+    }
+
+    function it_logs(LoggerInterface $logger, Normalizer $normalizer, RequestInterface $request, ResponseInterface $response)
```
sagikazarmark: I think this begs the question: logs...what?

spec/Normalizer/NormalizerSpec.php
```diff
@@ -0,0 +1,36 @@
...
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\UriInterface;
+use Psr\Log\LoggerInterface;
+
+class NormalizerSpec extends ObjectBehavior
+{
+    function it_is_initializable(LoggerInterface $logger)
+    {
+        $this->beAnInstanceOf('Http\Client\Plugin\Normalizer\Normalizer');
```
sagikazarmark: shouldHaveType, beAnInstanceOf is a construction thing.

spec/Normalizer/NormalizerSpec.php
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    function it_normalize_request_to_string(RequestInterface $request)
+    {
+        $request->getMethod()->shouldBeCalled()->willReturn('GET');
```
sagikazarmark: shouldBeCalled is only necessary when the behavior is the fact that a method is called. For example you expect wait to be called on the promise. But in these cases, the behavior is the value being returned, not the fact that the method is called.

src/LoggerPlugin.php
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Log request, response and exception for a HTTP Client
+ *
+ * @author Joel Wurtz <joel.wurtz@gmail.com>
+ */
+class LoggerPlugin implements Plugin
+{
+    /**
+     * @var LoggerInterface Logger to log request / response / exception for a http call
```
sagikazarmark: Please place the comment above the type declaration.

src/LoggerPlugin.php
```diff
@@ -0,0 +1,74 @@
...
+     */
+    private $logger;
+
+    /**
+     * @var Normalizer Normalize request and response to string or array
+     */
+    private $normalizer;
+
+    public function __construct(LoggerInterface $logger, Normalizer $normalizer)
```
sagikazarmark: Please add docblock for the constructor with typehints (description not required).

src/Normalizer/Normalizer.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace Http\Client\Plugin\Normalizer;
+
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * Normalize a request or a response into a string or an array
```
sagikazarmark, line 9: Author info

src/Normalizer/Normalizer.php
```diff
@@ -0,0 +1,41 @@
...
+     * @return string
+     *
+     * @internal
+     */
+    public function normalizeResponseToString(ResponseInterface $response)
+    {
+        return sprintf("%s %s %s", $response->getStatusCode(), $response->getReasonPhrase(), $response->getProtocolVersion());
+    }
+}
+
```
sagikazarmark: Missing newline

src/Normalizer/Normalizer.php
dbu, line 9: can you please also explain in the class docblock that this is `@internal` and should not be used outside this package?

joelwurtz, line 9: Hum i add `@internal` on the method, but this normalizer is injected into the plugin, to allow user to change how to render the log line, so it's not that "internal" ?

sagikazarmark, line 9: It is...for now. We should replace it in the future with some kind of Formatter interface with different kind of formatters.

dbu, line 9: oh, i see. hm. it does make sense to inject the string converter. but it will be a BC break when we start a separate tool and switch to that. could we for now simply instantiate Normalizer in the plugin and not have an argument for Normalizer? then we can add an optional argument once we have built the separate package to convert requests to strings, without creating a BC break for users.

joelwurtz:
I have moved normalizer to `@internal` should be good to merge.

sagikazarmark:
Nice job @joelwurtz

PR #9, opened 2015-11-13
Decoder and ContentLength plugin

joelwurtz:
Decoder plugin allows to decode gzip, deflate, compress or chunked body from the response
ContentLengthPlugin "sanitize" the content length header, by setting its value if possible, and if not, (like we are streaming from a process), decorate the body with a chunkstream (so it's compliant with the RFC where no content length header is set)


.travis.yml
```diff
@@ -19,7 +19,7 @@ env:
 
 matrix:
   allow_failures:
-    - php: 7.0
+    - php: hhvm
```
dbu: i would rather skip those tests on hhvm. and have the plugin throw an exception in the constructor if it is incompatible with hhvm

src/DecoderPlugin.php
```diff
@@ -0,0 +1,134 @@
...
+        if ($this->useContentEncoding) {
+            $request = $request->withHeader('Accept-Encoding', ['gzip', 'deflate', 'compress']);
+        }
+
+        return $next($request)->then(function (ResponseInterface $response) {
+            return $this->decodeResponse($response);
+        }, function (Exception $exception) use($request) {
+            if ($exception instanceof Exception\HttpException) {
+                $response  = $this->decodeResponse($exception->getResponse());
+                $exception = new Exception\HttpException($exception->getMessage(), $request, $response, $exception);
```
dbu: do we need to do something to not lose the exact exception in this case? or is HttpException final?

src/DecoderPlugin.php
```diff
@@ -0,0 +1,134 @@
...
+                $stream = $this->decorateStream($encoding, $response->getBody());
+
+                if (false === $stream) {
+                    array_unshift($newEncodings, $encoding);
+
+                    continue;
+                }
+
+                $response = $response->withBody($stream);
+            }
```
dbu, line 96: duplication alert. can we move this to a private method instead?

src/DecoderPlugin.php
```diff
@@ -0,0 +1,134 @@
...
+
+            $response = $response->withHeader('Transfer-Encoding', $newEncodings);
+        }
+
+        if ($this->useContentEncoding && $response->hasHeader('Content-Encoding')) {
```
dbu: can a response have both Transfer-Encoding and Content-Encoding? do we really want to follow both headers if both are set? (i don't know, just a bit worried of side effects. and the order matters here.

src/DecoderPlugin.php
```diff
@@ -0,0 +1,134 @@
...
+class DecoderPlugin implements Plugin
+{
+    /**
+     * @var bool Whether this plugin decode stream with value in the Content-Encoding header (default to true).
+     *
+     * If set to false only the Transfer-Encoding header will be used.
+     */
+    private $useContentEncoding;
+
+    public function __construct($useContentEncoding = true)
```
dbu: i think we should have the long description of useContentEncoding as docblock on the constructor as this is what people doing `new DecoderPlugin` will see.

src/DecoderPlugin.php
sagikazarmark: No, it is not. Client and Server error exceptions should extend HttpException.

dbu:
this is really cool! do clients like guzzle re-implement all that inside their code? on the other hand, what will guzzle do itself in the end? not much?

src/DecoderPlugin.php
dbu: then we should make sure we create the same exception that we caught, right?

joelwurtz: Yeah but constructor may not be the same :/

joelwurtz:
Guzzle doesn't implement this, but i think curl does (only for the transfer encoding header if my memory is good)

src/DecoderPlugin.php
joelwurtz: Yes traditionally you don't have gzip, deflate, compress in the transfer encoding but in the content encoding, transfer encoding is used for the message part not the entity part. 

The difference is mainly for user need, as someone may went to store the gzip content and not the decoded one (think of a reverse proxy like varnish with a vary on the Accept Encoding by example)

src/DecoderPlugin.php
dbu: hm, tricky that one then. i think degrading the exception to a less specific one is unexpected and a problem. maybe we should ignore execptions and say that this plugin must come after the Error plugin so that the response is decoded when thrown? probably makes sense in general to do the encoding very late.

joelwurtz: Or if the HttpException as a setResponse method it would also work

joelwurtz: Or wondering if it's possible to make the constructor final ?

sagikazarmark: Exception should be immutable I think. I don't see any blocker why the constructor could not be immutable. Although I agree with @dbu: order of extensions matter and should be the user's responsibility.

dbu: > Although I agree with @dbu: order of extensions matter and should be the
> user's responsibility.

mark: so you agree that say this plugin should simply ignore exception?

sagikazarmark: Agree, the exception conversion should either be the last step after every other plugin ran or the first and fail without any further plugin processing.

joelwurtz: >  fail without any further plugin processing

Impossible by design, since we don't even know if we got an exception or not when adding behavior.

Should i just remove handling of the response in the exception, anyways CookiePlugin does not care about the exception it should be consistent so yeah removing this.

.travis.yml
joelwurtz: I'm more in the thinking of providing no support for hhvm, i really dont know anyone who is using it, and with upcoming php7 it may be even worse. 
IMO I don't want to pass too many hours supporting something not used at all (and rather use them to write more things for php-http :) ).

sagikazarmark: I would say HHVM is not a top priority for now, but it would be nice to support it eventually.

dbu: i have no worries making the tests skip things around streams + hhvm. but i would like to keep the hhvm build green. at liip we do use it in some projects, for example. skipping tests should not be a big hassle. if we end up with most things not working on hhvm, we can reconsider.

src/DecoderPlugin.php
joelwurtz, line 96: Removed

joelwurtz:
Will work on hhvm implementation,  aside from this, any other remarks ? or is this good to merge ?

sagikazarmark:
I haven't really followed this thread, so ask @dbu for any remarks. :+1: from me, awesome work @joelwurtz as usual.

dbu:
looks good to me. i suggest that you simply add something to the tests to skip. if it would be phpunit, it would be

``` php
public function setUp()
{
     if (defined('HHVM_VERSION')) {
         $this->markTestSkipped();
    }
}
```

sagikazarmark:
IIRC Akeneo has a SkipExtension for PHPSpec.

dbu:
if its too much hassle, i can add a conditional `rm` of the tests in the
.travis.yml files to simply remove the tests when running with hhvm.

sagikazarmark:
> if its too much hassle

AFAIK it uses the same logic/annotations as phpunit does. How is it too much hassle?

However I checked it and it can only check existing classes/interfaces, not the PHP env.

dbu:
> However I checked it and it can only check existing classes/interfaces, not the PHP env.

thats what i meant with too much hassle. if we need to find some phpspec plugins and whatnot, we can just as well do it from .travis.yml. as we agree hhvm is not top priority, we should have a simple way to skip those tests.

sagikazarmark:
I would still stick to the setUp/let method instead of some weird rm logic. :stuck_out_tongue_winking_eye:

sagikazarmark:
``` php
use PhpSpec\Exception\Example\SkippingException;
//...
function let()
{
    if(defined('HHVM_VERSION')) {
        throw new SkippingException();
    }
}
```

dbu:
ah, if that works then sure lets do that. why is that namespace "Example"?

sagikazarmark:
Because tests are called "examples" in phpspec terminology:
- There are Suites which are collections of tests (for example you have one spec/ directory)
- There are Specifications which are collections of examples (one class, several test methods)
- There are Examples which are the actual test methods (eg. it_does_something)

joelwurtz:
Ok on hhvm, ready to merge ?

dbu:
great. can you please move hhvm out of the allowed to fail tests again? then we are good to merge.

we have to remember to mention in the plugin doc that the encoding plugin does not support hhvm.

dbu:
actually, maybe you can mark the stream library as conflicting with hhvm? we have no hard dependencies on the libraries used by the plugins, so people trying to run this will notice there is a problem even if they don't read the documentation.

sagikazarmark:
Conflict means uninstallable, right?

dbu:
it means composer require / install will fail when running under hhvm, yes.

i am not suggesting to make plugin conflict with hhvm, only the stream
library joel wrote

joelwurtz:
test will failed then on hhvm for this, as we require this lib in dev packages, any way of avoiding that ?

dbu:
oh well. i guess that won't work. lets leave it like this then. just move hhvm out of allowed failures and then we merge this.

we could remove it from require-dev and have travis run composer require if not on hhvm, but that is complicated. i guess hhvm users are used to having to check whether libraries work on hhvm...

joelwurtz:
Encoding library can work on hhvm, just requiring implementing many filters from the php world :/

dbu:
:+1:

PR #15, opened 2015-11-27
Added a PSR6 cache plugin

Nyholm:
Added a draft on a cache plugin. This can not be merged until PSR6 is final (currently voting). 
- [x] require-dev psr/cache
- [x] write unit tests

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,7 @@
         "php-http/cookie": "^0.1@dev",
         "symfony/stopwatch": "^2.3",
         "psr/log": "^1.0",
+        "psr/cache": "1.0.0",
         "php-http/encoding": "^0.1@dev"
     },
     "autoload": {
@@ -34,6 +35,7 @@
         "php-http/cookie": "Allow to use CookiePlugin",
         "symfony/stopwatch": "Allow to use the StopwatchPlugin",
         "psr/log-implementation": "Allow to use the LoggerPlugin",
+        "psr/cache-implementation": "Allow to use the CachePlugin",
         "php-http/encoding": "Allow to use the Decoder and Encoder plugin"
     },
     "scripts": {
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -1,0 +1,168 @@
+<?php
+
+namespace Http\Client\Plugin;
+
+use Http\Client\Tools\Promise\FulfilledPromise;
+use Psr\Cache\CacheItemPoolInterface;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * Allow for caching a response.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class CachePlugin implements Plugin
+{
+    /**
+     * @var CacheItemPoolInterface
+     */
+    private $pool;
+
+    /**
+     * Default time to store object in cache. This value is used if CachePlugin::respectCacheHeaders is false or
+     * if cache headers are missing.
+     *
+     * @var int
+     */
+    private $defaultTtl;
+
+    /**
+     * Look at the cache headers to know how long this response is going to be cached.
+     *
+     * @var bool
+     */
+    private $respectCacheHeaders;
+
+    /**
+     * @param CacheItemPoolInterface $pool
+     * @param array                  $options
+     */
+    public function __construct(CacheItemPoolInterface $pool, array $options = [])
+    {
+        $this->pool = $pool;
+        $this->defaultTtl = isset($options['default_ttl']) ? $options['default_ttl'] : null;
+        $this->respectCacheHeaders = isset($options['respect_cache_headers']) ? $options['respect_cache_headers'] : true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
+    {
+        $method = strtoupper($request->getMethod());
+
+        // if the request not is cachable, move to $next
+        if ($method !== 'GET' && $method !== 'HEAD') {
+            return $next($request);
+        }
+
+        // If we can cache the request
+        $key = $this->createCacheKey($request);
+        $cacheItem = $this->pool->getItem($key);
+
+        if ($cacheItem->isHit()) {
+            // return cached response
+            return new FulfilledPromise($cacheItem->get());
+        }
+
+        return $next($request)->then(function (ResponseInterface $response) use ($cacheItem) {
+            if ($this->isCacheable($response)) {
+                $cacheItem->set($response)
+                    ->expiresAfter($this->getMaxAge($response));
+                $this->pool->save($cacheItem);
+            }
+
+            return $response;
+        });
+    }
+
+    /**
+     * Verify that we can cache this response.
+     *
+     * @param ResponseInterface $response
+     *
+     * @return bool
+     */
+    protected function isCacheable(ResponseInterface $response)
+    {
+        if (!in_array($response->getStatusCode(), [200, 203, 300, 301, 302, 404, 410])) {
+            return false;
+        }
+        if ($this->getCacheControlDirective($response, 'no-store') || $this->getCacheControlDirective($response, 'private')) {
+            return false;
+        }
+
+        return true;
+    }
+
+    /**
+     * Returns the value of a parameter in the cache control header. If not found we return false. If found with no
+     * value return true.
+     *
+     * @param ResponseInterface $response
+     * @param string            $name
+     *
+     * @return bool|string
+     */
+    private function getCacheControlDirective(ResponseInterface $response, $name)
+    {
+        $headers = $response->getHeader('Cache-Control');
+        foreach ($headers as $header) {
+            if (preg_match(sprintf('|%s=?([0-9]+)?|i', $name), $header, $matches)) {
+
+                // return the value for $name if it exists
+                if (isset($matches[1])) {
+                    return $matches[1];
+                }
+
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * @param RequestInterface $request
+     *
+     * @return string
+     */
+    private function createCacheKey(RequestInterface $request)
+    {
+        return md5($request->getMethod().' '.$request->getUri());
+    }
+
+    /**
+     * Get a ttl in seconds. It could return null if we do not respect cache headers and got no defaultTtl.
+     *
+     * @param ResponseInterface $response
+     *
+     * @return int|null
+     */
+    private function getMaxAge(ResponseInterface $response)
+    {
+        if (!$this->respectCacheHeaders) {
+            return $this->defaultTtl;
+        }
+
+        // check for max age in the Cache-Control header
+        $maxAge = $this->getCacheControlDirective($response, 'max-age');
+        if (!is_bool($maxAge)) {
+            $ageHeaders = $response->getHeader('Age');
+            foreach ($ageHeaders as $age) {
+                return $maxAge - ((int) $age);
+            }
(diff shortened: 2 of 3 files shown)


src/CachePlugin.php
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     * We can only cache 2* responses.
+     *
+     * @param ResponseInterface $response
+     *
+     * @return bool
+     */
+    private function isCacheableResponse(ResponseInterface $response)
+    {
+        return floor($response->getStatusCode() / 100) === 2;
```
dbu: i think $statusCode >= 200 && $statusCode < 300 would be more readable. symfony is doing this:

https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/HttpFoundation/Response.php#L521-L529

i think we should respect cache control headers from the server.

src/CachePlugin.php
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     * @param ResponseInterface $response
+     *
+     * @return CacheItemInterface
+     */
+    private function createCacheItem(ResponseInterface $response)
+    {
+        $ttl = $this->getMaxAge($response);
+
+        // TODO instantiate an implementation of CacheItemInterface properly.
```
dbu: if i read https://github.com/php-fig/fig-standards/blob/master/proposed/cache.md#cacheiteminterface correctly, we are supposed to talk to the pool to get the instance, no?

it would be sad if we need to depend on a concrete implementation in the end.

dbu:
looks cool! before we merge, i think you should add the psr cache to require-dev also and add some unit tests.

Nyholm:
Thank you for the feedback. I agree and I will make some changes. 

I have not added psr/cache to the require dev because it is not on packagist yet.

src/CachePlugin.php
```diff
@@ -0,0 +1,162 @@
...
+     * @var bool
+     */
+    private $respectCacheHeaders;
+
+    /**
+     * @param CacheItemPoolInterface $pool
+     * @param int                    $defaultTtl
+     * @param bool                   $respectCacheHeaders
+     */
+    public function __construct(CacheItemPoolInterface $pool, $defaultTtl = 60, $respectCacheHeaders = true)
```
sagikazarmark: Not sure, but shouldn't this be multiline? (PSR)

src/CachePlugin.php
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * @param RequestInterface $request
+     * @param callable         $next
+     * @param callable         $first
+     *
+     * @return mixed
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
```
sagikazarmark, line 51: Inheritdoc

src/CachePlugin.php
```diff
@@ -0,0 +1,162 @@
...
+    {
+        if (!$this->isCacheableRequest($request)) {
+            return $next($request);
+        }
+
+        // If we can cache the request
+        $key = $this->createCacheKey($request);
+        $cacheItem = $this->pool->getItem($key);
+        if ($cacheItem->exists()) {
```
sagikazarmark: Newline before if

src/CachePlugin.php
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * We do only cache GET and HEAD requests.
+     *
+     * @param RequestInterface $request
+     *
+     * @return bool
+     */
+    private function isCacheableRequest(RequestInterface $request)
```
sagikazarmark: isCacheable, request comes after anyway.

sagikazarmark:
Nice one @Nyholm, thanks.

A few more ideas:
- Cache Control has Shared Max Age as well, although I am not sure we should care about it
- Do we really need all those oneliners (checks mainly) in separate methods? Can they be reused?
- Are there any possibility that we will have more options in the future? If yes, I think an options array in the constructor would be better.

src/CachePlugin.php
Nyholm: Only if more that 80 chars if I remember correctly.

src/CachePlugin.php
Nyholm: Are you suggesting that I should rename `isCacheableRequest` to `isCacheable`. How can I make a difference between `isCacheableRequest` and `isCacheableResponse`?

dbu:
> I have not added psr/cache to the require dev because it is not on
> packagist yet.

ah sorry, makes sense :-) i will update the PR description with a
tasklist to keep an overview

src/CachePlugin.php
dbu: +1 for a $options array with `default_ttl` and `respect_cache_headers` options for now. will be easier to scale.

dbu:
i would not use s-maxage i think. writing a proxy in PHP sounds like a weird idea. people could still extend the plugin if they found a use case for this and overwrite the getMaxAge method. or do a pull request for an option to use s-maxage instead of max-age.

Nyholm:
Thank you for the feedback. 

I have reduced the number of small one liner functions and improved the `getCacheControlDirective`.

About the Shared Max Age: I think (not sure) that shared max age should not be used by browser or clients. Only by reversed proxies. 

About using array in the constructor: Is this forward compatible? Wouldn't a configuration object be better?

sagikazarmark:
> About using array in the constructor: Is this forward compatible? Wouldn't a configuration object be better?

I don't think we need a configuration object. An array can be extended dynamically without BC breaks, new options can be documented.

Not sure if we need it, just mentioned the possibility.

dbu:
we could use the OptionsResolver from symfony would allow to validate the configuration. could be a nice compromise. a value object for configuration is a lot of effort and a bit awkward e.g. in symfony.

Nyholm:
Okey, I have added some tests and make sure to use an array as second constructor argument to be more extendable in the future. 
The tests will fail because of missing dependencies for psr/cache.

joelwurtz:
What about etag and receving 304 Not modified ?

Nyholm:
Do we want to support 304 or etag? If we do, we still have to make a HTTP request. The win is that we can get a small request back with a 304 header if the resource is "Not modified".

dbu:
If-None-Match / If-Not-Modified headers would be really cool i think. for example a client doing some polling on a server that supports etags / last-modified headers.

i was thinking if we can add that separately, but i think its tightly coupled - you need the cached answer for the 304 to make any sense. the user of HTTPlug will still want to receive the answer. so ideally we just have an option if validation should be attempted - i guess it can default to true, its part of the HTTP spec.

in terms of caching, the tricky bit will be to keep answers cached for longer than max-age says, if they have an etag/last-modified header. basically we could keep them indefinitely. maybe we could have an optional config option for a maximum cache lifetime.

btw, note that if both last-modified _and_ etag are present, BOTH have to be validated (in the sense of an AND, not OR).

Nyholm:
I've updated the PR. The composer.json includes the git repository for psr/cache. Once the voting is complete we could use packagist instead.

src/CachePlugin.php
```diff
@@ -0,0 +1,167 @@
...
+            return $next($request);
+        }
+
+        // If we can cache the request
+        $key = $this->createCacheKey($request);
+        $cacheItem = $this->pool->getItem($key);
+
+        if ($cacheItem->isHit()) {
+            // return cached response
+            return $cacheItem->get();
```
joelwurtz: Hum, we save a response and so we return a reponse (when hitting the cache), but this method must return a promise, don't think it will work.

Should it not be a `new FulfilledPromise($cacheItem->get())` ?

src/CachePlugin.php
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    /**
+     * Verify that we can cache this response.
+     *
+     * @param ResponseInterface $response
+     *
+     * @return bool
+     */
+    private function isCacheable(ResponseInterface $response)
```
joelwurtz: Can we have this protected ? So if someone want to have its own rule for caching a reponse it is possible without rewriting everything.

src/CachePlugin.php
```diff
@@ -0,0 +1,167 @@
...
+     * @var bool
+     */
+    private $respectCacheHeaders;
+
+    /**
+     * @param CacheItemPoolInterface $pool
+     * @param array                  $options
+     */
+    public function __construct(CacheItemPoolInterface $pool, array $options = array())
+    {
```
joelwurtz, line 42: Not a fan of having an options array for only 2 paramters, can we not use direct parameter in the constructor ?

src/CachePlugin.php
Nyholm: Sure, Good idea

src/CachePlugin.php
Nyholm, line 42: We mentioned it [here](https://github.com/php-http/plugins/pull/15#discussion_r46047604) and I changed to an array.

sagikazarmark, line 42: See the discussion in the PR.

src/CachePlugin.php
Nyholm: You are correct. Thank you for finding this issue!

Nyholm:
Thank you for the feedback @joelwurtz. I've updated the PR now.

src/CachePlugin.php
joelwurtz, line 42: Yeah but expect for having less characters in the constructor is there another reason @dbu ?

Having options array is useful for dynamic keys / parameters, however in this case we loose the potential of type hinting and make things more confuse for the user.

sagikazarmark, line 42: Not just less characters. Less number of parameters. Not just now, but in the future as well.

> we loose the potential of type hinting

Adding additional arguments is not forbidden. Just config-like options are groupped.

joelwurtz, line 42: Yes, but if a config like option is an instance of an object ?

Don't really mind to have those kinds of options array, but in my experience if they are not well checked (with something like symfony/options resolver for example) they often lead to many bugs and user misunderstanding.

dbu, line 42: we depend on the symfony OptionsResolver already. i would propose to emply the resolver here to do validation.

the plus side of the array is that adding more optional configuration with default value will be simply with an array, and much trickier if they all need an explicit order. (for BC, parameter order can not change, so required parameters could end up after optional parameters and whatnot.

composer.json
```diff
@@ -10,6 +10,12 @@
             "email": "joel.wurtz@gmail.com"
         }
     ],
+    "repositories": [
+        {
+            "type": "git",
+            "url":  "https://github.com/php-fig/cache.git"
```
dbu: just to not forget: we need to remove this before we merge.

Nyholm: Yes, PSR-6 should be tagged and added to packagist this week.

src/CachePlugin.php
sagikazarmark, line 42: LOL, I've never thought about parameter order being a BC change.

dbu, line 42: > LOL, I've never thought about parameter order being a BC change.

when there are string arguments or other non-type hinted, they are
amongst the nastiest BC problems as you don't notice them immediately.

joelwurtz:
PSR6 available on packagist :), so just need need a update of the composer and it's good to merge IMO

sagikazarmark:
:+1:

Nyholm:
I've updated the PR. I see that StyleCI fails but it not because of my changes. 

I added #17 to fix the code style of the repo.

sagikazarmark:
Merged it into master, can you rebase, so that build can run here as well?

Nyholm:
Rebased!

sagikazarmark:
@Nyholm thanks

Nyholm:
Thank you for all the feedback and comments

PR #18, opened 2015-12-17
Add history plugin

joelwurtz:


dbu:
do we want to provide an in-memory implementation of the journal in here? or are people supposed to write their own implementation?

dbu:
nice and short, i like it!

sagikazarmark:
Maybe we could provide a Psr3 implementation?

Nyholm:
Not sure if that is a good idea. If we require a PSR3 implementation, how does this plugin differ from the [LoggerPlugin](https://github.com/php-http/plugins/blob/master/src/LoggerPlugin.php)?

dbu:
i guess the main idea of history is to do tests or other analysis during
run time. everything else is likely about logging. or about caching maybe.

sagikazarmark:
Okay, makes sense. In that case an in-memory journal would probably make sense. I imagine it could be something like the BatchResponse in utils.

joelwurtz:
In fact, this is mainly to interface this with some frameworks, or other debug tools.

Don't know if it's useful to provide an implementation ? Use case i have in mind here is to have a class which use this interface and bridge DataCollector from symfony to have request / response / exceptions in the debug toolbar.

sagikazarmark:
@joelwurtz cool idea, like that. In fact, I would really need a solution like that RIGHT NOW. :stuck_out_tongue: 

But I also think that it might be useful for testing as @dbu said.

joelwurtz:
We can merge this as it, and do a in memory implementation later :)

sagikazarmark:
Fine by me. Maybe you could add an issue to document this plugin as well.

joelwurtz:
Yes, need to add an issue for the cache plugin documentation also

PR #19, opened 2015-12-18
Dont respect cache headers when $this->respectCacheHeaders=false

Nyholm:


src/CachePlugin.php
```diff
@@ -89,7 +89,9 @@ protected function isCacheable(ResponseInterface $response)
         if (!in_array($response->getStatusCode(), [200, 203, 300, 301, 302, 404, 410])) {
             return false;
         }
-        if ($this->getCacheControlDirective($response, 'no-store') || $this->getCacheControlDirective($response, 'private')) {
+        if ($this->respectCacheHeaders &&
```
sagikazarmark: I think that this check could be in one return statement.

Nyholm:
I've updated the PR according to your comments

sagikazarmark:
Actually I meant the first condition as well. :smile: 

You could set some helper variables (like the status code list).

sagikazarmark:
`($this->getCacheControlDirective($response, 'no-store') || $this->getCacheControlDirective($response, 'private'))`

This could go into a separate method, like `checkCacheControlCacheble` or `checkCacheControlAllowsCache` or something like that.

Nyholm:
ah. Okey. =)

I'll update again and make sure it is readable.

Nyholm:
hm, okey. I put it in a variable now since it not reused.

src/CachePlugin.php
```diff
@@ -86,14 +86,11 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
-        }
-        if ($this->getCacheControlDirective($response, 'no-store') || $this->getCacheControlDirective($response, 'private')) {
-            return false;
-        }
+        $cachableCodes = [200, 203, 300, 301, 302, 404, 410];
+        $privateHeaders = $this->getCacheControlDirective($response, 'no-store') || $this->getCacheControlDirective($response, 'private');
 
-        return true;
+        // If http status code is cachable and if we respect the headers, make sure there is no private cache headers.
+        return in_array($response->getStatusCode(), $cachableCodes) && (!$this->respectCacheHeaders || !$privateHeaders);
```
sagikazarmark: Hm...

Isn't it the same as `!($this->respectCacheHeaders && $privateHeaders)`

Nyholm: It is the same according to De Morgan's law. You are correct.

sagikazarmark: > De Morgan's law

I knew that there is a name for this, but I couldn't recall.

sagikazarmark:
Thanks @Nyholm

Nyholm:
Thank you for merging.

src/CachePlugin.php
```diff
@@ -86,14 +86,11 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
-        }
-        if ($this->getCacheControlDirective($response, 'no-store') || $this->getCacheControlDirective($response, 'private')) {
-            return false;
-        }
+        $cachableCodes = [200, 203, 300, 301, 302, 404, 410];
+        $privateHeaders = $this->getCacheControlDirective($response, 'no-store') || $this->getCacheControlDirective($response, 'private');
 
-        return true;
+        // If http status code is cachable and if we respect the headers, make sure there is no private cache headers.
+        return in_array($response->getStatusCode(), $cachableCodes) && !($this->respectCacheHeaders && $privateHeaders);
```
dbu, line 93: i think the previous code with early returns was more readable. i will do another small PR :-)

dbu, line 93: see #22

Issue #20, opened 2015-12-18
[CachePlugin] Streams are not cached


Nyholm:
I found an issue with the CachePlugin. When we retrieve a `Response` form the cache we have lost the body because the stream was not in cache. 

``` php
if ($response->getBody()->__toString() === '') {
  // Always true
}
```

Im not sure how to solve this in a good way... One solution is to recreate the stream when we fetch the data from cache. This require us to add a dependency on `php-http/discovery`.

``` php
// If we can cache the request
$key = $this->createCacheKey($request);
$cacheItem = $this->pool->getItem($key);

if ($cacheItem->isHit()) {
     // return cached response
     $data = $cacheItem->get();
     $response = $data['response'];
     $response->withBody(StreamFactoryDiscovery::find()->createStream($data['body']));
     return new FulfilledPromise($response);
}

return $next($request)->then(function (ResponseInterface $response) use ($cacheItem) {
     if ($this->isCacheable($response)) {
         $cacheItem->set(['response'=>$response, 'body'=>$response->getBody()->__toString()])
              ->expiresAfter($this->getMaxAge($response));
              $this->pool->save($cacheItem);
      }

      return $response;
});

```

sagikazarmark:
> This require us to add a dependency on `php-http/discovery`.

I would prefer adding the message-factory package as a dependency and inject it in the constructor.

Nyholm:
Okey, Thank you for the feedback. I'll send a PR.

Nyholm:
Closed by #21

PR #21, opened 2015-12-18
[CachePlugin] Streams are not cached

Nyholm:
This PR solves two issues. First issue #20 where streams are not cached. Secondly we add a OptionsResolver to validate the configuration.

### DIFF
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -7,6 +7,7 @@ use Http\Message\StreamFactory;
 use Psr\Cache\CacheItemPoolInterface;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
  * Allow for caching a response.
@@ -26,35 +27,27 @@ class CachePlugin implements Plugin
     private $streamFactory;
 
     /**
-     * Default time to store object in cache. This value is used if CachePlugin::respectCacheHeaders is false or
-     * if cache headers are missing.
-     *
-     * @var int
+     * @var array
      */
-    private $defaultTtl;
-
-    /**
-     * Look at the cache headers to know whether this response may be cached and to 
-     * decide how it can be cached.
-     *
-     * @var bool Defaults to true
-     */
-    private $respectCacheHeaders;
+    private $config;
 
     /**
      * Available options are
      *  - respect_cache_headers: Whether to look at the cache directives or ignore them.
-     * 
+     *  - default_ttl: If we do not respect cache headers or can't calculate a good ttl, use this value.
+     *
      * @param CacheItemPoolInterface $pool
      * @param StreamFactory          $streamFactory
-     * @param array                  $options
+     * @param array                  $config
      */
-    public function __construct(CacheItemPoolInterface $pool, StreamFactory $streamFactory, array $options = [])
+    public function __construct(CacheItemPoolInterface $pool, StreamFactory $streamFactory, array $config = [])
     {
         $this->pool = $pool;
         $this->streamFactory = $streamFactory;
-        $this->defaultTtl = isset($options['default_ttl']) ? $options['default_ttl'] : null;
-        $this->respectCacheHeaders = isset($options['respect_cache_headers']) ? $options['respect_cache_headers'] : true;
+
+        $optionsResolver = new OptionsResolver();
+        $this->configureOptions($optionsResolver);
+        $this->config = $optionsResolver->resolve($config);
     }
 
     /**
@@ -105,7 +98,7 @@ class CachePlugin implements Plugin
         if (!in_array($response->getStatusCode(), [200, 203, 300, 301, 302, 404, 410])) {
             return false;
         }
-        if (!$this->respectCacheHeaders) {
+        if (!$this->config['respect_cache_headers']) {
             return true;
         }
         if ($this->getCacheControlDirective($response, 'no-store') || $this->getCacheControlDirective($response, 'private')) {
@@ -160,8 +153,8 @@ class CachePlugin implements Plugin
      */
     private function getMaxAge(ResponseInterface $response)
     {
-        if (!$this->respectCacheHeaders) {
-            return $this->defaultTtl;
+        if (!$this->config['respect_cache_headers']) {
+            return $this->config['default_ttl'];
         }
 
         // check for max age in the Cache-Control header
@@ -181,6 +174,24 @@ class CachePlugin implements Plugin
             return (new \DateTime($header))->getTimestamp() - (new \DateTime())->getTimestamp();
         }
 
-        return $this->defaultTtl;
+        return $this->config['default_ttl'];
+    }
+
+    /**
+     * Configure an options resolver.
+     *
+     * @param OptionsResolver $resolver
+     *
+     * @return array
+     */
+    private function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver->setDefaults([
+            'default_ttl' => null,
+            'respect_cache_headers' => true,
+        ]);
+
+        $resolver->setAllowedTypes('default_ttl', ['int', 'null']);
+        $resolver->setAllowedTypes('respect_cache_headers', 'bool');
     }
 }
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -77,7 +77,15 @@ class CachePlugin implements Plugin
 
         return $next($request)->then(function (ResponseInterface $response) use ($cacheItem) {
             if ($this->isCacheable($response)) {
-                $cacheItem->set(['response' => $response, 'body' => $response->getBody()->__toString()])
+                $bodyStream = $response->getBody();
+                $body = $bodyStream->__toString();
+                if ($bodyStream->isSeekable()) {
+                    $bodyStream->rewind();
+                } else {
+                    $response = $response->withBody($this->streamFactory->createStream($body));
+                }
+
+                $cacheItem->set(['response' => $response, 'body' => $body])
                     ->expiresAfter($this->getMaxAge($response));
                 $this->pool->save($cacheItem);
             }
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -189,8 +189,6 @@ class CachePlugin implements Plugin
      * Configure an options resolver.
      *
      * @param OptionsResolver $resolver
-     *
-     * @return array
      */
     private function configureOptions(OptionsResolver $resolver)
     {
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -3,6 +3,7 @@
 namespace Http\Client\Plugin;
 
 use Http\Client\Tools\Promise\FulfilledPromise;
+use Http\Message\StreamFactory;
 use Psr\Cache\CacheItemPoolInterface;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
@@ -18,6 +19,11 @@ class CachePlugin implements Plugin
      * @var CacheItemPoolInterface
      */
     private $pool;
+
+    /**
+     * @var StreamFactory
+     */
+    private $streamFactory;
 
     /**
      * Default time to store object in cache. This value is used if CachePlugin::respectCacheHeaders is false or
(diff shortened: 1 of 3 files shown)


src/CachePlugin.php
```diff
@@ -36,11 +42,13 @@ class CachePlugin implements Plugin
 
     /**
      * @param CacheItemPoolInterface $pool
+     * @param StreamFactory          $streamFactory
      * @param array                  $options
      */
-    public function __construct(CacheItemPoolInterface $pool, array $options = [])
+    public function __construct(CacheItemPoolInterface $pool, StreamFactory $streamFactory, array $options = [])
```
sagikazarmark: I think this is a great use case why options should probably moved out of the constructor and configured by a setter.

src/CachePlugin.php
```diff
@@ -36,11 +42,13 @@ class CachePlugin implements Plugin
...
     {
         $this->pool = $pool;
+        $this->streamFactory = $streamFactory;
         $this->defaultTtl = isset($options['default_ttl']) ? $options['default_ttl'] : null;
```
sagikazarmark: I know this is not a current change, but shouldn't we use options resolver here?

sagikazarmark:
Looks good to me, but I am not as involved in plugins as I would like to, so I would wait for others feedback.

src/CachePlugin.php
Nyholm: Yeah, we talked about that before. I could implement it.

src/CachePlugin.php
Nyholm: Tell me more, why would you configure mandatory dependencies with setters? That means that the CachePlugin object will be i an unusable state before you use that setter.

sagikazarmark: Ah, no, you misunderstood me. I only meant the options parameter, which is not mandatory by definition.

Nyholm: Ah, okey, sure. 

If we should not have the options array in the constructor, I would prefer to use multiple setter for each property: 

``` php

    /**
     * @param boolean $respectCacheHeaders
     *
     * @return CachePlugin
     */
    public function setRespectCacheHeaders($respectCacheHeaders)
    {
        $this->respectCacheHeaders = $respectCacheHeaders;

        return $this;
    }

    /**
     * @param int $defaultTtl
     *
     * @return CachePlugin
     */
    public function setDefaultTtl($defaultTtl)
    {
        $this->defaultTtl = $defaultTtl;

        return $this;
    }
```

sagikazarmark: Wouldn't it kill the idea behind the options array and options resolver? Or did we reach the maximum number of possible options?

Nyholm: Yes, you would miss the validation from the option resolver. But you would not have the issue with default values and misstyped array keys.

sagikazarmark: I have no problem with having setters for options if we can be confident that we don't need to add tons of setters later.

Nyholm:
Ive updated the PR to use option resolver. I make sure to be consistent with [SocketHttpClient](https://github.com/php-http/socket-client/blob/master/src/SocketHttpClient.php)

src/CachePlugin.php
```diff
@@ -20,29 +22,26 @@ class CachePlugin implements Plugin
...
+    ];
 
     /**
      * @param CacheItemPoolInterface $pool
-     * @param array                  $options
+     * @param StreamFactory          $streamFactory
+     * @param array                  $config
      */
-    public function __construct(CacheItemPoolInterface $pool, array $options = [])
+    public function __construct(CacheItemPoolInterface $pool, StreamFactory $streamFactory, array $config = [])
```
sagikazarmark, line 43: What's the problem with options? :stuck_out_tongue:

src/CachePlugin.php
```diff
@@ -20,29 +22,26 @@ class CachePlugin implements Plugin
...
     {
         $this->pool = $pool;
-        $this->defaultTtl = isset($options['default_ttl']) ? $options['default_ttl'] : null;
-        $this->respectCacheHeaders = isset($options['respect_cache_headers']) ? $options['respect_cache_headers'] : true;
+        $this->streamFactory = $streamFactory;
+
+        $this->config = $this->configure($config);
```
sagikazarmark: The reason why a setter would be better (no matter if it is per options or just one accepting an array) is that we don't break BC if we need a typehinted parameter. Config/options is optional anyway.

src/CachePlugin.php
```diff
@@ -160,6 +163,22 @@ private function getMaxAge(ResponseInterface $response)
...
+
+    /**
+     * @param array $config
+     *
+     * @return array
+     */
+    protected function configure(array $config = [])
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults($this->config);
```
sagikazarmark: IMO you shouldn't rely on the object's state when setting configuration defaults. What if you want to configure it again?

src/CachePlugin.php
Nyholm, line 43: None, but SocketHttpClient was using `$config` so I thought I would be consistent.

src/CachePlugin.php
Nyholm: When the config array is in the constructor, there is no issue. But I'll change this when I move it a setter. 
Thank you for the feedback.

src/CachePlugin.php
sagikazarmark, line 43: Okay, makes sense

src/CachePlugin.php
dbu: i would prefer to have a `private function getOptionsResolver` that prepares the resolver with our defaults and the constructor assigns the resolve() result to $this->config.

not sure whether there is some sort of best practice around this - there are a couple of ways to do it. could also be all in the constructor.

sagikazarmark: I think the best practice is to have a configureOptions method which accepts an options resolver object. You can create that in the constructor, configure it with the configure options and after that you can either use it in the constructor or later with setters.

dbu: sounds good to me. but i would keep that method private, no need to
customize that. and assign directly in the constructor. changing plugin
config at runtime would feel odd to me

sagikazarmark: :+1: private, agree

Nyholm:
I've created a `configureOptions` to configure the option resolver. You can set config by both a constructor argument and a setter.

Nyholm:
> Before Symfony 2.6, setAllowedTypes() and addAllowedTypes() expected the values to be given as an array mapping option names to allowed types: $resolver->setAllowedTypes(array('port' => array('null', 'int')));

They do not any more. This means I should update the dependency on options resolver to 2.6 which means we are dropping support for Symfony 2.3. 

I could maybe do some if(method_exists()) to find out what version we currently using. Is that a way we want to go?

dbu:
imo support for symfony 2.8 and 3 is enough. people that don't even update symfony inside the 2.x versions won't introduce php-http into their code i think.

sagikazarmark:
Agree with @dbu

Nyholm:
Okey. I've updated to 2.6. There is no reason to bump composer.json to 2.8 because we dont use any features introduced after 2.6. (Im not sure if there are any.. )

src/CachePlugin.php
```diff
@@ -169,6 +180,34 @@ private function getMaxAge(ResponseInterface $response)
...
+    }
+
+    /**
+     * Set config to the plugin. This will overwrite any previously set config values.
+     *
+     * @param array $config
+     */
+    public function setConfig(array $config)
+    {
+        $this->config = $this->optionsResolver->resolve($config);
```
dbu: i would inline that in the constructor. there is no need to change the configuration at runtime. and then we can make the OptionsResolver a local class.

Nyholm: Okey, I though you wanted a setter for the config. Ive updated now

Nyholm:
The tests are failing because of https://github.com/php-http/encoding/commit/6a7a0162d8b72534ca6c906d8c9f81954e7e529f

Nyholm:
Ping @joelwurtz, What do you think about this?

src/CachePlugin.php
```diff
@@ -67,12 +68,16 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
 
         if ($cacheItem->isHit()) {
             // return cached response
-            return new FulfilledPromise($cacheItem->get());
+            $data = $cacheItem->get();
+            $response = $data['response'];
+            $response = $response->withBody($this->streamFactory->createStream($data['body']));
```
dbu, line 73: does it make sense that the body is always a stream? should we keep what it was when we had a cache miss?

src/CachePlugin.php
```diff
@@ -67,12 +68,16 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
+
+            return new FulfilledPromise($response);
         }
 
         return $next($request)->then(function (ResponseInterface $response) use ($cacheItem) {
             if ($this->isCacheable($response)) {
-                $cacheItem->set($response)
+                $cacheItem->set(['response' => $response, 'body' => $response->getBody()->__toString()])
```
joelwurtz: We need to have some sort of buffered stream, as some implementations provide a unbuffered stream, so reading a stream once will certainly make it not available afterwards.

Also even if the stream is buffered by the implementation, it must be rewind here.

Concretly we must add something like, if stream is seekable just rewind this, if not buffer it into a new stream.

We may face memory problem if big file are downloaded, but don't know if we can avoid that

dbu: what happens with a stream body when setting the cache? what happens without a stream body? duplicated data in cache? maybe we could check if the body is a stream and if so set it to the cache item, otherwise only set the response? and when restoring check if we have a body in the cache and if we do create the stream from that?

dbu: > We may face memory problem if big file are downloaded, but don't know if
> we can avoid that

sounds like a warning for the cache plugin doc: if you handle large
downloads, make sure the implementation works as expected. or contribute
to make it better.

src/CachePlugin.php
Nyholm, line 73: Yes, The body has to be a stream, always. That is part of PSR-7.

sagikazarmark, line 73: Actually, I think it can be null as well, can't it?

Nyholm, line 73: Hm, I might be wrong. But `getBody` should [always return a stream](https://github.com/php-fig/http-message/blob/master/src/MessageInterface.php#L169).

sagikazarmark, line 73: You are right. Initially, it was the plan to return null when the body is empty. All these PSR7 versions in my head sometimes make me wonder about the REAL interfaces.

src/CachePlugin.php
Nyholm: Thank you for the comments. 

> Concretly we must add something like, if stream is seekable just rewind this, if not buffer it into a new stream.

Hm, okey. I think I understand what you mean.. Give me a minute and I'll update

> > We may face memory problem if big file are downloaded, but don't know if we can avoid that
> 
> sounds like a warning for the cache plugin doc: if you handle large
> downloads, make sure the implementation works as expected. or contribute
> to make it better. 

Agreed.

src/CachePlugin.php
```diff
@@ -67,12 +68,24 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
-                $cacheItem->set($response)
+                $bodyStream = $response->getBody();
+                $body = $bodyStream->__toString();
+                if ($bodyStream->isSeekable()) {
+                    $bodyStream->rewind();
+                } else {
+                    $response = $response->withBody($this->streamFactory->createStream($body));
+                }
+
+                $cacheItem->set(['response' => $response, 'body' => $body])
```
dbu, line 88: is the body field that is a stream simply ignored when storing in the cache? could a cache implementation have a problem with the stream?

dbu:
:+1:

src/CachePlugin.php
Nyholm, line 88: The cache implementation have no problem storing a stream object. But the stream object may wrap php//temp which wont be available when the cached object is retrieved. 

So yes, the content of the stream is ignored by the cache if we try to store a stream object.

src/CachePlugin.php
```diff
@@ -169,6 +182,24 @@ private function getMaxAge(ResponseInterface $response)
...
-        return $this->defaultTtl;
+        return $this->config['default_ttl'];
+    }
+
+    /**
+     * Configure an options resolver.
+     *
+     * @param OptionsResolver $resolver
+     *
+     * @return array
```
sagikazarmark: It does not return anything.

src/CachePlugin.php
sagikazarmark, line 88: How awesome would it be if we could write a stream to the cache and the read it as a stream back from the cache. I guess it would require some kind of special backend which immediately limits our possibilities in the term of interoperability. It is still a good idea. :wink:

Nyholm, line 88: Yeah, it would be cool =)

joelwurtz:
LGTM, Thanks @Nyholm

Nyholm:
Thank you for merging!

PR #22, opened 2015-12-19
improve isCacheable readability

dbu:
seems we have different ideas of code style. i think this is much more readable than what we did in #19

wdyt?

### DIFF
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -28,13 +28,17 @@ class CachePlugin implements Plugin
     private $defaultTtl;
 
     /**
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Well, I don't mind after all.

sagikazarmark:
@dbu do you have access to the repositories in StyleCI?

dbu:
fixed the styling. seems that the github online editor is not that convenient for clean style...

sagikazarmark:
:smile: 

At least it turned out that you do have access to StyleCI which is cool.

Nyholm:
:+1:

PR #27, opened 2015-12-22
Let the normalizer how schema and host

Nyholm:
Reading my log files when using the LoggerPlugin I can see a entry like this: 

```
[2015-12-22 19:18:50] app.INFO: Emit request: "GET /language/translate/v2?key=param 1.1" {"request":"[object] (GuzzleHttp\\Psr7\\Request: {})"} []
[2015-12-22 19:18:50] app.INFO: Receive response: "200 OK 1.1" for request: "GET /language/translate/v2?key=param 1.1" {"request":"[object] (GuzzleHttp\\Psr7\\Request: {})","response":"[object] (GuzzleHttp\\Psr7\\Response: {})"} []
```

I would like to see to what host I was requesting data from. 

```
// Current
GET /language/translate/v2?key=param 1.1

// Suggested
GET https://www.googleapis.com/language/translate/v2?key=param 1.1
```

This PR makes sure I can see the host and schema in the log file.

### DIFF
--- a/src/Normalizer/Normalizer.php
+++ b/src/Normalizer/Normalizer.php
@@ -23,7 +23,7 @@ class Normalizer
      */
     public function normalizeRequestToString(RequestInterface $request)
     {
-        return sprintf('%s %s %s', $request->getMethod(), $request->getRequestTarget(), $request->getProtocolVersion());
+        return sprintf('%s %s %s', $request->getMethod(), $request->getUri()->__toString(), $request->getProtocolVersion());
     }
 
     /**
--- a/spec/LoggerPluginSpec.php
+++ b/spec/LoggerPluginSpec.php
@@ -10,6 +10,7 @@ use PhpSpec\ObjectBehavior;
 use Prophecy\Argument;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\UriInterface;
 use Psr\Log\LoggerInterface;
 
 class LoggerPluginSpec extends ObjectBehavior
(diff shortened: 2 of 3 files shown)


src/Normalizer/Normalizer.php
```diff
@@ -23,7 +23,7 @@ class Normalizer
      */
     public function normalizeRequestToString(RequestInterface $request)
     {
-        return sprintf('%s %s %s', $request->getMethod(), $request->getRequestTarget(), $request->getProtocolVersion());
+        return sprintf('%s %s %s', $request->getMethod(), $request->getUri()->__toString(), $request->getProtocolVersion());
```
sagikazarmark, line 26: I think `getRequestTarget` is what you want to use here.

sagikazarmark:
No objections from me. Although the full uri might not always contain scheme and host. These things can be in header as well, can't they?

Nyholm:
The `schema` is never in the header. The `host` must be in the header on version 1.1 but not 1.0. 

`getRequestTarget` will return the request target which is the second word on the first line of the HTTP message. This will never include host or schema. 

`$request->getUri->__toString()` will let the implementation to deal with problems to put together the URL.

sagikazarmark:
Ah, you are right.

I wonder why the FIG chose to use the `__toString` method instead of a `build` or something like that. I get that this way you can cast the object to string. But internally it doesn't make sense: you can just call the build method. Anyone else can do whatever they want with their object.

It's like having a constructor in a trait.

Nyholm:
Question: Say that I am terrible unhappy with the current normalizer. I would like to be able to use my own. Should I let the user inject a normalizer from in the constructor? Or that might be another PR?

Use case: Say I don't want my API keys in the log so I want to 'wash' those away before logging.

sagikazarmark:
Hm, I have a feeling that these normalizers should go into the new message package:

https://github.com/php-http/message

@joelwurtz @dbu what do you think?

Nyholm:
> Hm, I have a feeling that these normalizers should go into the new message package: https://github.com/php-http/message

I feel that would be another PR, lets do this change first.

dbu:
+1 to merge as is.

as to having custom normalizers: we discussed the idea of a normalizer that outputs a line that you can copy into a shell and call `curl` (like chrome and firefox dev tools allow you to do). so yes, we should eventually make the normalizer injectable and probably have an interface for it. i guess we can do that at the same time as we move it from here to message.

sagikazarmark:
Thanks @Nyholm

PR #29, opened 2015-12-25
Replace normalizer with message formatter

sagikazarmark:


sagikazarmark:
ping @Nyholm

Nyholm:
One thing I would like to have in my logs are the response body when the response fails. It often have some detailed error messages. That means I would create a formatter like this: 

``` php
public function formatResponse(ResponseInterface $response)
    {
        return sprintf(
            '%s %s %s with body %s',
            $response->getStatusCode(),
            $response->getReasonPhrase(),
            $response->getProtocolVersion(),
            $response->getBody()->__toString()
        );
    }
```

In my error logs I would get a line like: 

`Error: "Bad request" with response: "400, Bad request 1.1 with body [Big and awfully long body. ]" when emitting request: "GET /something"`

I think I would like the response to always be at the end of the log entry.

So make the log entry look like this: 

``` php
 sprintf('Error: "%s" with when emitting request: "%s". This is the response %s', $exception->getMessage(), $this->formatter->formatRequest($request), $this->formatter->formatResponse($exception->getResponse())),
```

sagikazarmark:
You can. There is an interface in the message package which allows you to implement your own formatter.

Nyholm:
Sorry, not done with the post... one sec

Nyholm:
I was too fast on the keys and accidentally posted the comment before I was done writing it.

joelwurtz:
+1 With @Nyholm so having response or not does not change the beginning of the message : 

With response : 

```
sprintf('Error: "%s" when emitting request: "%s" with the response %s', $exception->getMessage(), $this->formatter->formatRequest($request), $this->formatter->formatResponse($exception->getResponse()))
```

Without : 

```
sprintf('Error: "%s" when emitting request: "%s"', $exception->getMessage(), $this->formatter->formatRequest($request));
```

Is that ok with you @Nyholm ?

Nyholm:
> Is that ok with you @Nyholm ?

Yes, That is what I meant. 

I think this PR is a great improvement. One could argue that you should be able to configure the complete log entry but I think that is overkill. 

Good job @sagikazarmark!

sagikazarmark:
This PR is strictly about replacing Normalizer with formatter, also introduces an interface and possibility to use custom formatters. If you would like to change the log message, please do it in a separate PR/issue.

sagikazarmark:
Thanks @Nyholm

joelwurtz:
We could always make the write of log entry / errors in decoupled protected function so someone can extend this class and change the message. But yeah another PR.

sagikazarmark:
I think this is the same case as with the error plugin: it is not that complicated that someone cannot write his/her own. We created the plugin system for this purpose after all. We agree on a common use case and provide that. If someone has different needs, it is not that hard to implement a custom plugin.

In HttplugBundle plugin services could even be replace with a custom one.

PR #31, opened 2015-12-25
Add a safety catch for max restarts

### LINKED ISSUE #7: safety catch
@dbu
could we have a magic "first" plugin that is inserted by the PluginClient that counts calls and prevents infinite restarts?

@joelwurtz
Not a fan of doing this as the boundaring between infinite restart and many redirection request can be thin sometimes.

@sagikazarmark
This could be a good feature I think, but it should not be automatically added. Can be good in a "fire&forget" situation, where you want to limit the number of restarts.

@dbu
maybe have an option to define how many restarts should be allowed?
we could also just accept a plugin for that, but thats very implicit and
not happening by default. and imho this is something that should happen
by default to protect people not aware of the risk. endless redirects
are pretty simple to create if the server is doing something stupid
(like redirect to the same url you are already on)

@sagikazarmark
An option sounds good to me, but it should be optional. I could imagine a setter for this option value, but the constructor should probably be limited to what we already have there.

@joelwurtz
Actually any plugin using the $first callable or doing retry must implement this mechanism (it's already done in the RetryPlugin and RedirectPlugin) we can maybe provide a SafePluginClient implementation (which extends the current one) to add this behavior ?

@dbu
i would see an `$options` argument to the PluginClient that allows to overwrite a default value (which could be 10 or 20 or something even larger). the thing is that the architecture is prone to endless loops. some people won't read documentation and end up with too many nested calls or even worse things when they run into this with a custom plugin. having something on by default that can be tweaked / disabled sounds better to me.

joelwurtz:
See #7

### DIFF
--- a/spec/PluginClientSpec.php
+++ b/spec/PluginClientSpec.php
@@ -4,6 +4,7 @@ namespace spec\Http\Client\Plugin;
 
 use Http\Client\HttpAsyncClient;
 use Http\Client\HttpClient;
+use Http\Client\Plugin\Plugin;
 use Http\Promise\Promise;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
@@ -44,5 +45,20 @@ class PluginClientSpec extends ObjectBehavior
 
         $this->beConstructedWith($asyncClient);
         $this->sendAsyncRequest($request)->shouldReturn($promise);
+    }
+
+    function it_throws_loop_exception(HttpClient $client, RequestInterface $request)
+    {
+        $this->beConstructedWith($client, [new DefectuousPlugin()]);
+
+        $this->shouldThrow('Http\Client\Plugin\Exception\LoopException')->duringSendRequest($request);
+    }
+}
+
+class DefectuousPlugin implements Plugin
+{
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
+    {
+        return $first($request);
     }
 }
--- a/spec/PluginClientSpec.php
+++ b/spec/PluginClientSpec.php
@@ -10,6 +10,14 @@ use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use PhpSpec\ObjectBehavior;
 
+class DefectuousPlugin implements Plugin
+{
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
+    {
+        return $first($request);
+    }
+}
+
 class PluginClientSpec extends ObjectBehavior
 {
     function let(HttpClient $client)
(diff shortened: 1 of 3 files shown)


joelwurtz:
Hum failure is weird, is this a PHP 7 bug ?

spec/PluginClientSpec.php
```diff
@@ -45,4 +46,19 @@ function it_sends_async_request_with_underlying_client(HttpAsyncClient $asyncCli
...
+    {
+        $this->beConstructedWith($client, [new DefectuousPlugin()]);
+
+        $this->shouldThrow('Http\Client\Plugin\Exception\LoopException')->duringSendRequest($request);
+    }
+}
+
+class DefectuousPlugin implements Plugin
+{
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
```
sagikazarmark: Try using the FQCN here. Might be a PHP 7, but. This workaround could temporarily solve it. Don't forget adding a warning, that we had to make our hands dirty because of that.

joelwurtz: Will see, currently installing php7 :)

sagikazarmark: Using phpbrew or custom build?

joelwurtz: Using docker

sagikazarmark:
Have you found a solution?

I did: moving the stub into a separate file solves the problem.

joelwurtz:
Yes just see that my fix wasn't fixing xD, it is good in my new PR

joelwurtz:
I think this is due to the way phpspec work by changing the file.

src/PluginClient.php
```diff
@@ -87,6 +117,17 @@ private function createPluginChain($pluginList)
             $firstCallable = $lastCallable;
         }
 
+        $firstCalls = 0;
+        $firstCallable = function (RequestInterface $request) use ($options, $lastCallable, &$firstCalls) {
```
dbu, line 121: do we need the local $options? i thought since php 5.4 you can use $this inside the anonymous function.

src/PluginClient.php
```diff
@@ -28,12 +30,20 @@ class PluginClient implements HttpClient, HttpAsyncClient
     protected $plugins;
 
     /**
+     * A list of options.
+     *
+     * @var array
+     */
+    protected $options;
+
+    /**
```
dbu, line 39: we should list and explain the options (in the main docblock text of the constructor, not in the argument)

dbu:
:+1:

src/PluginClient.php
sagikazarmark, line 121: :+1:

Issue #14, opened 2015-11-23
Do not use sendAsyncRequest for synchronous requests


mekras:
PluginClient uses sendAsyncRequest method for synchronous requests. This may lead to overhead and slowdown, because async logic is more complex.

dbu:
i think the reason was to avoid duplicated code. but maybe we can find a way to avoid too much duplication? @joelwurtz would it be enough to create a different $lastCallable depending on what was requested?

joelwurtz:
Maybe the lastest callable can be changed but we have to keep promise
structure
Le 23 nov. 2015 4:19 PM, "David Buchmann" notifications@github.com a
écrit :

sagikazarmark:
I have to agree. Also, if not sticking to promises, we would still have to build a callable chain which would probably end up with the same complexity.

dbu:
agree, we would not want to end up with duplicated code for every plugin. so the EmulateAsyncClient would be what @mekras asks for, right? or where you worried by something else than the actual client call?

sagikazarmark:
Actually, now that all of the logic is moved from the EmulateAsync client to client-tools package, we could move back the EmulateAsyncClient class here and remove the utils dependency.

PR #33, opened 2015-12-25
Prefer the sync call if possible

joelwurtz:
See #14


src/PluginClient.php
```diff
@@ -62,9 +65,20 @@ public function __construct($client, array $plugins = [], array $options = [])
      */
     public function sendRequest(RequestInterface $request)
     {
-        $promise = $this->sendAsyncRequest($request);
+        if (!($this->client instanceof HttpClient)) {
```
sagikazarmark, line 70: Shouldn't we just emulate a sync client if an async client is passed?

src/PluginClient.php
```diff
@@ -62,9 +65,20 @@ public function __construct($client, array $plugins = [], array $options = [])
...
+            return $this->sendAsyncRequest($request)->wait();
+        }
+
+        $client = $this->client;
+        $pluginChain = $this->createPluginChain($this->plugins, function (RequestInterface $request) use ($client) {
```
sagikazarmark, line 77: Technically, this doesn't make any difference IMO. When a non-async client is passed, it is emulated to be an async client, which means you are free to always rely on sendAsyncRequest. If it is an async client, the upper condition would just return with an async client call. If it is not an async client, then the emulated async request send method will just do the same as this callback does.

What's the difference then?

src/PluginClient.php
```diff
@@ -72,7 +86,10 @@ public function sendRequest(RequestInterface $request)
      */
     public function sendAsyncRequest(RequestInterface $request)
     {
-        $pluginChain = $this->createPluginChain($this->plugins);
+        $client = $this->client;
+        $pluginChain = $this->createPluginChain($this->plugins, function (RequestInterface $request) use ($client) {
```
sagikazarmark, line 94: In PHP 5.4, you can access the `$this` variable inside the closure.

sagikazarmark:
See php-http/client-common#2

src/PluginClient.php
joelwurtz, line 77: The difference is if you have an async client and a sync client, it will use the correct underlying method for sendRequest or sendAsyncRequest, before it was always using the sendAsyncRequest one.

src/PluginClient.php
joelwurtz, line 70: Yes we can, but by doing this we save many calls, from sync to async and async to sync.

sagikazarmark, line 70: It doesn't matter IMO. If we emulate a sync client as well, this condition and call becomes unnecessary.

src/PluginClient.php
sagikazarmark, line 77: Yeah, but I still can't see the difference. Here is what happens now:

If the client is async (not nonasync) then the async version is called.

If it is sync, we emulate an async one. In the plugin clien's sync call we wrap the sync client call with the same code as we did in case of the emulated client. Same code, whats the difference apart from the method name?

src/PluginClient.php
joelwurtz, line 70: The emulation add a non necessary layer of code in this case. 

As we will have a emulated sync call, which transform promise into a response / exception which is then transformed into promise, and then retransformed into response / exception.

Here we only do the conversion from promise to response / exception once.

sagikazarmark, line 70: Sorry, I meant async emulation here. We don't need sync emulation IMO. There you also have only one conversion (like now).

src/PluginClient.php
joelwurtz, line 77: And if it is both async and sync ?

dbu:
:+1:

Issue #36, opened 2015-12-26
Make psr/log and psr/cache hard dependencies


sagikazarmark:
Those are light, interface packages, probably installed in most projects anyway.

joelwurtz:
:-1: Making this a hard dependancy will force user to have a log and cache implementation

sagikazarmark:
That's not true. I meant the interface packages, not the virtual packages.

joelwurtz:
The interface package does not require a virtual package ?

sagikazarmark:
Nope. And that's how it should be, like on case of HTTPlug: it doesn't require the virtual packages either.

joelwurtz:
Oh ok, sounds good to me then

PR #37, opened 2015-12-26
Dependencies

sagikazarmark:


joelwurtz:
:+1: To merge if nothing fails

sagikazarmark:
#36 should be here if it gets accepted.

PR #38, opened 2015-12-28
Add initial documentation

sagikazarmark:


docs/implement-your-own.md
```diff
@@ -0,0 +1,92 @@
...
+When writing your own Plugin, you need to be aware that the Plugin Client is async first.
+This means that every plugin must be written with Promises. More about this later.
+
+Each plugin must implement the `Http\Client\Plugin\Plugin` interface.
+
+This interface defines the `handleRequest` method that allows to modify behavior of the call:
+
+```php
+/**
+ * Handles the request and returns the response coming from the next callable.
```
xabbuh, line 12: Is this actually true? I mean a plugin could also replace the response coming from the next callable (for example, to filter it), couldn't it?

docs/implement-your-own.md
```diff
@@ -0,0 +1,92 @@
...
+public function handleRequest(RequestInterface $request, callable $next, callable $first)
+{
+    $newRequest = $request->withHeader('MyHeader', 'MyValue');
+
+    return $next($newRequest);
+}
+```
+
+The `$first` callable is the first plugin in the chain. It allows you to completely reboot the execution chain, or send
+other request if needed, while still going through all the defined plugins.
```
xabbuh, line 45: [...] another request [...] or [...] other requests [...]

docs/implement-your-own.md
```diff
@@ -0,0 +1,92 @@
...
+Like in case of the `$next` callable, you must pass the `$request` as the first argument.
+
+```
+public function handleRequest(RequestInterface $request, callable $next, callable $first)
+{
+    if ($someCondition) {
+        $newRequest = new Request();
+        $promise = $first($newRequest);
+
+        // Use the promise do some jobs ...
```
xabbuh, line 55: [...] promise to do [...]

docs/implement-your-own.md
```diff
@@ -0,0 +1,92 @@
...
+    return $next($request);
+}
+```
+
+!!! warning "Warning:"
+    In this example the condition is not superfluous:
+    you need to have some way to not call the `$first` callable each time
+    or you will end up in an infinite execution loop.
+
+The `$next` and `$first` callable will return a Promise (defined in `php-http/promise`).
```
xabbuh, line 67: callables

docs/implement-your-own.md
xabbuh, line 12: Oh, you also do something like that (modifying the response) in one example below.

docs/implement-your-own.md
```diff
@@ -0,0 +1,92 @@
...
+        echo $exception->getMessage();
+
+        throw $exception;
+    });
+}
+```
+
+!!! warning "Warning:"
+    Contract for the `Http\Promise\Promise` is temporary until
+    [PSR is released](https://groups.google.com/forum/?fromgroups#!topic/php-fig/wzQWpLvNSjs).
```
xabbuh, line 87: [...] until a PSR [...]

docs/index.md
```diff
@@ -0,0 +1,79 @@
...
+        $redirectPlugin,
+    ]
+);
+```
+
+The Plugin Client accepts and implements both `Http\Client\HttpClient` and `Http\Client\HttpAsyncClient`, so you can use
+both ways to send a request. In case the passed client implements only one of these interfaces, the Plugin Client
+"emulates" the other behavior as a fallback.
+
+It is important, that the order of plugins matter. During the request, plugins are called in the order they have
```
xabbuh, line 52: matters

docs/index.md
```diff
@@ -0,0 +1,79 @@
...
+been added, from first to last. Once a response has been received, they are called again in a reversed order,
```
xabbuh, line 53: [...] in reversed order [...]

docs/index.md
```diff
@@ -0,0 +1,79 @@
...
+
+In case of our previous example, the execution chain will look like this:
+
+```
+Request  ---> PluginClient ---> RetryPlugin ---> RedirectPlugin ---> HttpClient ----
+                                                                                   | (processing call)
+Response <--- PluginClient <--- RetryPlugin <--- RedirectPlugin <--- HttpClient <---
+```
+
+In order to have correct behavior over the global process, you need to understand well how each plugin used,
```
xabbuh, line 64: [...] plugin is used [...]

sagikazarmark:
@xabbuh Thanks for the feedback. This was just a PoC PR to see how documentation in plugins repo would work.

I will apply your fixes and move the stuff back to the main documentation.

PR #40, opened 2015-12-29
private properties and final PluginClient class

xabbuh:


xabbuh:
This also inlines `$this` into anonymous function callbacks (see @dbu's comment in https://github.com/php-http/plugins/pull/31#discussion_r48528603).

sagikazarmark:
:+1:

sagikazarmark:
Thanks @xabbuh

PR #42, opened 2015-12-29
AddHostPlugin

Nyholm:
This will implement baseUri functionality. It copies Guzzle's behaviour when the input starts with a slash. 

// base uri  is http://foo.com/api/

| Input | Output |
| --- | --- |
| http://baz.com/biz | http://baz.com/biz |
| foo | http://foo.com/api/foo |
| /foo | http://foo.com/foo |

### DIFF
--- a/src/AddHostPlugin.php
+++ b/src/AddHostPlugin.php
@@ -1,0 +1,73 @@
+<?php
+
+namespace Http\Client\Plugin;
+
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\UriInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * Add schema and host to a request. Can be set to overwrite the schema and host if desired.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class AddHostPlugin implements Plugin
+{
+    /**
+     * @var UriInterface
+     */
+    private $host;
+
+    /**
+     * @var bool
+     */
+    private $replace;
+
+    /**
+     * @param UriInterface $host
+     * @param array        $config {
+     *
+     *     @var bool $replace True will replace all hosts, false will only add host when none is specified.
+     * }
+     */
+    public function __construct(UriInterface $host, array $config = [])
+    {
+        if ($host->getHost() === '') {
+            throw new \LogicException('Host can not be empty');
+        }
+
+        $this->host = $host;
+
+        $resolver = new OptionsResolver();
+        $this->configureOptions($resolver);
+        $options = $resolver->resolve($config);
+
+        $this->replace = $options['replace'];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
+    {
+        if ($this->replace || $request->getUri()->getHost() === '') {
+            $uri = $request->getUri()->withHost($this->host->getHost());
+            $uri = $uri->withScheme($this->host->getScheme());
+
+            $request = $request->withUri($uri);
+        }
+
+        return $next($request);
+    }
+
+    /**
+     * @param OptionsResolver $resolver
+     */
+    private function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver->setDefaults([
+            'replace' => false,
+        ]);
+        $resolver->setAllowedTypes('replace', 'bool');
+    }
+}
--- a/spec/AddHostPluginSpec.php
+++ b/spec/AddHostPluginSpec.php
@@ -1,0 +1,86 @@
+<?php
+
+namespace spec\Http\Client\Plugin;
+
+use Http\Message\StreamFactory;
+use Http\Message\UriFactory;
+use Http\Promise\FulfilledPromise;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\StreamInterface;
+use Psr\Http\Message\UriInterface;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+
+class AddHostPluginSpec extends ObjectBehavior
+{
+    function let(UriInterface $uri)
+    {
+        $this->beConstructedWith($uri);
+    }
+
+    function it_is_initializable(UriInterface $uri)
+    {
+        $uri->getHost()->shouldBeCalled()->willReturn('example.com');
+        $this->beConstructedWith($uri);
+        $this->shouldHaveType('Http\Client\Plugin\AddHostPlugin');
+    }
+
+    function it_is_a_plugin(UriInterface $uri)
+    {
+        $uri->getHost()->shouldBeCalled()->willReturn('example.com');
+        $this->beConstructedWith($uri);
+
+        $this->shouldImplement('Http\Client\Plugin\Plugin');
+    }
+
+    function it_adds_domain(
+        RequestInterface $request,
+        UriInterface $host,
+        UriInterface $uri
+    ) {
+        $host->getScheme()->shouldBeCalled()->willReturn('http://');
+        $host->getHost()->shouldBeCalled()->willReturn('example.com');
+
+        $request->getUri()->shouldBeCalled()->willReturn($uri);
+        $request->withUri($uri)->shouldBeCalled()->willReturn($request);
+
+        $uri->withScheme('http://')->shouldBeCalled()->willReturn($uri);
+        $uri->withHost('example.com')->shouldBeCalled()->willReturn($uri);
+        $uri->getHost()->shouldBeCalled()->willReturn('');
+
+        $this->beConstructedWith($host);
+        $this->handleRequest($request, function () {}, function () {});
+    }
+
+    function it_replaces_domain(
+        RequestInterface $request,
+        UriInterface $host,
+        UriInterface $uri
+    ) {
+        $host->getScheme()->shouldBeCalled()->willReturn('http://');
+        $host->getHost()->shouldBeCalled()->willReturn('example.com');
+
+        $request->getUri()->shouldBeCalled()->willReturn($uri);
+        $request->withUri($uri)->shouldBeCalled()->willReturn($request);
+
+        $uri->withScheme('http://')->shouldBeCalled()->willReturn($uri);
+        $uri->withHost('example.com')->shouldBeCalled()->willReturn($uri);
+
+
+        $this->beConstructedWith($host, ['replace'=>true]);
+        $this->handleRequest($request, function () {}, function () {});
+    }
+
+    function it_does_nothing_when_domain_exists(
+        RequestInterface $request,
+        UriInterface $host,
+        UriInterface $uri
+    ) {
+        $request->getUri()->shouldBeCalled()->willReturn($uri);
+        $uri->getHost()->shouldBeCalled()->willReturn('default.com');
+
+        $this->beConstructedWith($host);
+        $this->handleRequest($request, function () {}, function () {});
+    }
+}


joelwurtz:
Nice, do you think it's possible to remove the UriFactory dependancy and only accept a UriInterface for this Plugin ?

IMO would have a better reusability and less complexity for user experience

Nyholm:
Yeah, you are probably right. That would be possible. I agree that it wound be more reusable but I wound also decrease the developer experience. We just push the responsibility to convert a uri string to an object away from us. 

But I agree with your suggestion. I'll update shortly.

Nyholm:
There you go. No UriFactory needed!

joelwurtz:
> We just push the responsibility to convert a uri string to an object away from us.

Yes that's something that i'm considering by adding some sort of factory around plugins

src/BaseUriPlugin.php
```diff
@@ -0,0 +1,72 @@
...
+     * @param RequestInterface $request
+     * @param callable         $next    Next middleware in the chain, the request is passed as the first argument
+     * @param callable         $first   First middleware in the chain, used to to restart a request
+     *
+     * @return Promise
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
+    {
+        $uriString = $request->getUri()->__toString();
+        if (!strpos($uriString, '://')) {
```
joelwurtz: Not possible to provide a base uri without host ? (Like we just want to add /api path for our url and keep current host ?)

joelwurtz: Hum nevermind my comment didn't read well, instead it's not possible to change the host with the base uri ?

For me base uri has 2 roles : 
- Adding some sort of base path if needed
- Changing host if needed (very useful when in a developement environment or staging, so we can call a service on another url just by adding this plugin)

Nyholm: Hm, I've looked at Guzzle's implementation and copied their functionality. Which means that if we have the following: 

| Base path | Input | Output |
| --- | --- | --- |
| http://foo.com/api/ | http://baz.com/biz | http://baz.com/biz |
| http://foo.com/api/ | foo | http://foo.com/api/foo |
| http://foo.com/api/ | /foo | http://foo.com/foo |

Do I understand you correctly when you suggest the base path should always replace the available parts of the input?
It is a great idea but I can see a problem with that. How do you know what to replace?

| Base path | Input | Output |
| --- | --- | --- |
| http://foo.com/api/ | http://baz.com/biz/bar | http://foo.com/api/biz/bar, http://foo.com/api/bar or http://foo.com/api/? |

I believe that the one that writes the api client should use this plugin to allow users to replace the base path with whatever. 

I would suggest a ReplacePath plugin to implement your functionality.

dbu: i would expect a base uri plugin to only trigger if my uri has no domain in it, and never replace an explicitly specified domain. agree with tobias on this one. if there is a real use case to replace domains, we can do a plugin, but it seems like a rather weird edge case where we might better let people write a custom plugin to handle their case. as in replace only some domains or whatever.

@Nyholm another question on this line: what about an uri like `://example.com/foo`, so one that is missing just the protocol part? should we handle that? i tend to say no, but we might want to be explicit in the documentation on that.

joelwurtz: Don'tt think it's possible to have something like `://example.com/foo` since it's coming from an uri object. So it would be a defectuous implementation and not our part IMO

joelwurtz:
Hum after all i'm not sure about this i don't see the value of a plugin like this.

Maybe we should separate this in 2 plugins : 
- `PrefixPathPlugin`: to change the path after host, it will simply take a string and add it in every request call
- `BaseHostPlugin`: change / add host for a request (this plugin can have an option to only set host if not defined)

So the current functionallity can be achieved by a combination of this 2 plugins, also it allow better control for the user and a better understanding on what it's going on.

WDYT ?

Nyholm:
i like the idea is splitting it. But I also think we should be very close to Guzzles functionality and change if we really have a good reason for it. 

If we look at the LinkedIn API there are two endpoints. One to change a Oauth code for a token and one to make API requests. With guzzle I (API developer) can instansiate a client with a base uri like "https://api.linkedin.com/api/" and do all my request with ->get("resource"). 
When I'm changing code for a token I would do ->get("https://www.linkedin.com/api/token")
The point is that I (API developer) can choose when to use the base uri and when to not use it. 

A plugin like this does not, however, replace the functionality for **always** changing the host and prefixing the path.

joelwurtz:
With your exemple i understand more the use case, didn't think of that :)

However i think this can be achieve with options to this 2 plugins
- Only add host if no host present (in the BaseHost plugin)
- Only add prefix if no trailing slash in path (in the PrefixPath plugin)

src/BaseUriPlugin.php
dbu: @joelwurtz `://example.com/foo` is a valid uri afaik. you use that for example to include css/js/images on pages that are served both over http and https, to avoid security warnings by the browser. (though most situations https inside a http page would be fine - but not the other way round)

sagikazarmark: Isn't it `//example.com/foo`? At least I always ommited the colon.

dbu:
imho, we should not provide a plugin to rewrite urls. chances are high you are doing some very specific hackery when needing this, and in that situation you better write your own plugin tailered to the case you need to fix.

i would stick with the cases guzzle covers for what we provide, it seems to me the least astonishing behaviour.

sagikazarmark:
I think we should keep single responsibilities and provide plugins for the most common use-cases. If someone needs some more functionality, writting plugins is not that hard.

joelwurtz:
For me the BaseHost plugin is a must have for everyone and a very common use case, as when you develop something or are in production you want to use different host for your api.

So having a BaseHost plugin is something needed, also your staging host may be shared with others applications (or you local) so you want to add a prefix in the path in this case.

I think it would be very helpful to have this plugins given this use case, and by adding these 2 plugins with the corresponding options we can cover this use case and the guzzle one.

sagikazarmark:
Note:
http://foo.com/api/     foo     http://foo.com/api/foo

This scenario is not possible with PSR-7. The interface requires a starting / if no domain is added, which transforms foo to /foo.

dbu:
ah, good catch mark. i guess that makes this discussion simpler: whatever the base uri is is prepended to the path. the "foo" case is thus leftover in guzzle from their own request, or guzzle psr7 request is not complying with the spec. imho we should not cater to specific implementation glitches. if somebody needs to "absolutize" paths, they can do that in their own library, e.g. in a wrapper around the request factory - we did that for FOSHttpCache.

what we want to be sure is that the base uri does not end on `/`, to avoid google.com//foo situations.

src/BaseUriPlugin.php
joelwurtz: I think it's like @sagikazarmark said `//exemple.com/foo`

BTW guzzle and diactoros use parse_url as the parser for the string, so this string would recognize as the path : 

```
$ php -r "var_dump(parse_url('://exemple.com/test'));" 
array(1) {
  'path' =>
  string(19) "://exemple.com/test"
}

$ php -r "var_dump(parse_url('//exemple.com/test'));"
array(2) {
  'host' =>
  string(11) "exemple.com"
  'path' =>
  string(5) "/test"
}
```

dbu: ups, you are right. the correct format without schema is // not ://

so do we agree that the code is correct?

dbu:
@Nyholm do you have time to wrap this one up? is everything clear or do we have uncertainties about what can be supported / should be supported?

Nyholm:
> @Nyholm do you have time to wrap this one up? is everything clear or do we have uncertainties about what can be supported / should be supported?

I will do that this afternoon.

Nyholm:
So, what I have heard there are quite a few scenarios we would like to solve. I've listed them below. Some facts: 
- We should not try to solve _everything_ in one plugin. 
- We can not copy the behavior form Guzzle because restrictions in PSR-7.

The use cases we want to have support for is (no particular order): 
A) `PrefixPathPlugin` - Adding/Prefixing a base path
B) `ReplaceHostPlugin` - Changing host (useful in development)
C) `BaseHostPlugin` - Adding host if none available

As in the role of a API client developer I would like to use a BaseHostPlugin for two reasons. (1) To avoid write the host all the time and (2) to allow users to change the host where I allow them by using a BaseHostPlugin themselves.  

As an application developer I would like to have the possibility replace **all** calls to production APIs. This is useful in staging environments to make sure we control where we send HTTP requests. 

Im not sure about the use case for a `PrefixPathPlugin`. @joelwurtz you might want to help here?

---

As an API client developer Im likely to write something like this: 

``` php
public function __construct(HttpClient $client = null) {
  if ($client === null) {
    $client = HttpClientDiscovery::find();
  }

  $this->client = new PluginClient($client, [BaseHostPlugin('http://example.com')]);
}

public function foo() {
  return $this->client->sendRequest(new Request('/api/resource'));
}
```

And when I use that API client I should do: 

``` php
$client = new PluginClient(new Http\Adapter\Guzzle6\Client(), [BaseHostPlugin('http://staging.com')]);
$api = new Api($client);
```

There is an issue here; http://staging.com will never be used because it belongs to the second plugin chain. At the time the second plugin chain is being evaluated there will always be a host. 

So instead I should do something like this: 

``` php
public function __construct(HttpClient $client = null) {
  if ($client === null) {
    $client = new PluginClient(HttpClientDiscovery::find(), [BaseHostPlugin('http://example.com')]);
  }

  $this->client = client;
}
```

... and then tell my users that they have to use the base plugin if they provide a new client. 

---

Would it make sense to have a `BaseHostPlugin` and a `ReplaceHostPlugin` or should their functionality be merged to one plugin?

``` php
class BaseHostPlugin {
  public function __construct($host, $replace = false) {}
}
```

joelwurtz:
i prefer the option way : BaseHostPlugin and ReplaceHostPlugin merged with an override option

> Im not sure about the use case for a PrefixPathPlugin. @joelwurtz you might want to help here?

It's the same use case as the ReplaceHostPlugin as maybe my dev environnement use subpath resolution instead of domain name resolution

joelwurtz:
It is also useful with a versioned api in the url, like /v1.1/articles, so people can write /articles and easily replace the version number with this plugin

Nyholm:
Are we talking about a prefixing, replacing and appending the URI?

sagikazarmark:
Not sure if it makes sense here, but there are some header plugins and those are prefixed with Header in the class name. Would it make sense to do the same here?

joelwurtz:
Only prefixing, also maybe we don't want to bother with that for the moment and think more about this use case in another PR / Issue

Nyholm:
Okey, I'll update this PR to be the `BaseHostPlugin` that will change the schema and domain. 
I think the default action should be "add if no host" and you have to choose if you want to force a replace. 

``` php
class BaseHostPlugin {
  public function __construct($host, $replace = false) {}
}
```

What do you think?

joelwurtz:
:+1: Please use an options array with symfony/options-resolver for the constructor as it's done in every plugin now

Nyholm:
> :+1: Please use an options array with symfony/options-resolver for the constructor as it's done in every plugin now

Thank you for the reminder. 

The PR is updated

src/BaseHostPlugin.php
```diff
@@ -0,0 +1,69 @@
...
+     * @var bool
+     */
+    private $replace;
+
+    /**
+     * Available options for $config are:
+     *  - replace: bool True will replace all hosts, false will only add host when none is specified.
+     *
+     * @param UriInterface $baseUri
+     * @param array        $config
```
joelwurtz: Look at https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#examples-12 on how the intent is for config array parameters in phpdoc, even if the standard is not yet accepted the format for config is pretty stable

joelwurtz:
:+1: with my comments :) Thanks @Nyholm

src/BaseHostPlugin.php
Nyholm: Yeah, I saw that but I choose not to include it in this PR because I wanted to be consistent with other plugins and then send a PR to fix them all. 

I'll make sure this plugin is doing it the correct way now already then. =)

joelwurtz:
LGTM any objection @sagikazarmark @dbu ?

spec/BaseHostPluginSpec.php
```diff
@@ -0,0 +1,83 @@
+<?php
+
+namespace spec\Http\Client\Plugin;
+
+use Http\Message\StreamFactory;
+use Http\Message\UriFactory;
+use Http\Promise\FulfilledPromise;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use Psr\Cache\CacheItemInterface;
```
sagikazarmark: What does PSR-6 do here?

spec/BaseHostPluginSpec.php
```diff
@@ -0,0 +1,83 @@
+<?php
+
+namespace spec\Http\Client\Plugin;
+
+use Http\Message\StreamFactory;
+use Http\Message\UriFactory;
+use Http\Promise\FulfilledPromise;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
```
sagikazarmark: I prefer havin test related imports at the end.

sagikazarmark:
None from me. Added some comments.

spec/BaseHostPluginSpec.php
Nyholm: Oops! Sorry.

src/BaseHostPlugin.php
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * @var bool
+     */
+    private $replace;
+
+    /**
+     * @param UriInterface $baseUri
+     * @param array        $config {
+     *     @var bool $replace True will replace all hosts, false will only add host when none is specified.
```
dbu: in the other plugins, i formatted this differently. yours is a good idea (though i would say `replace` not `$replace` as its an array key) but we should be consistent.

src/BaseHostPlugin.php
```diff
@@ -0,0 +1,68 @@
...
+     */
+    private $baseUri;
+
+    /**
+     * @var bool
+     */
+    private $replace;
+
+    /**
+     * @param UriInterface $baseUri
```
dbu: from a validation point of view it makes sense to use an object here, but we should call it $host and say that only the host part of the uri is used.

dbu: or we could simply go for a string.

src/BaseHostPlugin.php
```diff
@@ -0,0 +1,68 @@
...
+     * @param array        $config {
+     *     @var bool $replace True will replace all hosts, false will only add host when none is specified.
+     * }
+     */
+    public function __construct(UriInterface $baseUri, array $config = [])
+    {
+        $this->baseUri = $baseUri;
+
+        $resolver = new OptionsResolver();
+        $this->configureOptions($resolver);
```
dbu: i would inline this in the constructor, but thats really a matter of taste :-)

src/BaseHostPlugin.php
```diff
@@ -0,0 +1,68 @@
...
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\UriInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * Allow to modify the the schema and the host of an existing request.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class BaseHostPlugin implements Plugin
```
dbu: why "Base" ? maybe AddHostPlugin?

src/BaseHostPlugin.php
dbu: and should we call this replace or overwrite? maybe overwrite would be more explicit?

src/BaseHostPlugin.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+namespace Http\Client\Plugin;
+
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\UriInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * Allow to modify the the schema and the host of an existing request.
```
dbu: Add schema and host to a request. Can be set to overwrite the schema and host if desired.

src/BaseHostPlugin.php
dbu: sorry, lets forget about the string. it would mean we need the uri factory, or parse the host ourselves. not a good idea.

src/BaseHostPlugin.php
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * @param UriInterface $baseUri
+     * @param array        $config {
+     *     @var bool $replace True will replace all hosts, false will only add host when none is specified.
+     * }
+     */
+    public function __construct(UriInterface $baseUri, array $config = [])
+    {
+        $this->baseUri = $baseUri;
```
dbu: should we validate that this uri provides host and scheme?

src/BaseHostPlugin.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
+    {
+        if ($this->replace || $request->getUri()->getHost() === '') {
+            $uri = $request->getUri()->withHost($this->baseUri->getHost());
+            $uri = $uri->withScheme($this->baseUri->getScheme());
```
dbu: should we check for scheme separately? though i guess its a really weird case to have the scheme but not the host in a request. could it happen we have the host but not the scheme?

src/BaseHostPlugin.php
Nyholm: About the doc block formatting: https://github.com/php-http/plugins/pull/42#discussion_r50898065

Hm, replace vs overwrite. Im no native English speaker, but I think replace is more clear and describes the action better.

src/BaseHostPlugin.php
Nyholm: Yeah, I saw your changes on the other plugins. I was suggested to do like this on the CachePlugin and I feel it is easer to read this way. But as you say, it is a matter of taste. We use both ways at the moment.

src/BaseHostPlugin.php
Nyholm: Thank you

src/BaseHostPlugin.php
Nyholm: Base comes from "BaseUriPlugin". But AddHostPlugin may be better. It explains better the action of the plugin.

src/BaseHostPlugin.php
Nyholm: I do not think so. I may have a custom "AddSchemaIfConditionPlugin" later. If Im forced to use a schema here it may cause issue for my other plugin. I want this plugin to add what I tell it to add.

src/BaseHostPlugin.php
Nyholm: It does not really make sense. If you change the host you should not rely on the old schema.

dbu: okay.

src/AddHostPlugin.php
```diff
@@ -0,0 +1,68 @@
...
+     */
+    private $replace;
+
+    /**
+     * @param UriInterface $baseUri
+     * @param array        $config {
+     *     @var bool $replace True will replace all hosts, false will only add host when none is specified.
+     * }
+     */
+    public function __construct(UriInterface $baseUri, array $config = [])
```
dbu: sorry if you already answerd, but i can't find my question anymore. should we call this $host? baseUri was correct for the original way this was going, but now its different.

src/AddHostPlugin.php
```diff
@@ -0,0 +1,68 @@
...
+    {
+        $this->baseUri = $baseUri;
```
dbu: i still think verifying that scheme and host are set makes sense here. if we create other plugins for other purposes, they can have their own validation.

joelwurtz: Only host checking is necessary / required IMO, and since it's a immutable object :+1: for check in the constructor

src/AddHostPlugin.php
Nyholm: Ive must have missed that. Sorry.

src/AddHostPlugin.php
```diff
@@ -0,0 +1,72 @@
...
+     * @param UriInterface $host
+     * @param array        $config {
+     *     @var bool $replace True will replace all hosts, false will only add host when none is specified.
+     * }
+     */
+    public function __construct(UriInterface $host, array $config = [])
+    {
+        if ($host->getHost() === '') {
+            throw new \LogicException('Host can not be empty');
+        }
```
Nyholm, line 37: @dbu @joelwurtz Did you mean like this?

dbu, line 37: :+1:

src/AddHostPlugin.php
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
+    {
+        if ($this->replace || $request->getUri()->getHost() === '') {
+            $uri = $request->getUri()->withHost($this->host->getHost());
+            $uri = $uri->withScheme($this->host->getScheme());
```
dbu, line 55: what is $this->host->getScheme() is empty? does this set the scheme to the default scheme? we validate the host above, but not scheme.

joelwurtz, line 55: From the PSR7

> An empty scheme is equivalent to removing the scheme.

joelwurtz, line 55: So is this the wanted behavior, replace / remove existing if scheme is null (i think it's OK ?)

joelwurtz:
Could you fix the styleci issue like #51 ? and then go merge :)

Nyholm:
Updated

sagikazarmark:
Cool. Ready to merge and release?

Nyholm:
Yes

sagikazarmark:
@joelwurtz whenever you feel. :smile:

joelwurtz:
I'm feeling it right now ^^

joelwurtz:
Thanks @Nyholm !

Nyholm:
Thank you all for the feedback and discussions!

PR #44, opened 2016-01-17
Header Plugin

gsouf:
Adding `HeaderPlugin` that has the ability to set / remove headers.

``` php
// set the header 'foo: bar' if the header does not exist in the request
$headerPlugin->withHeader('foo', 'bar'); 

// set the header 'foo: bar' even if the header 'fooè already exists
$headerPlugin->withHeader('foo', 'bar', true);

// remove the header 'foo' if it exists on the request
$headerPlugin->withoutHeader('foo');
```

### DIFF
--- a/src/HeaderAppendPlugin.php
+++ b/src/HeaderAppendPlugin.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Http\Client\Plugin;
+
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Adds headers to the request.
+ * If the header already exists the value will be appended to the current value.
+ *
+ * This only makes sense for headers that can have multiple values like 'Forwarded'
+ *
+ * @link https://en.wikipedia.org/wiki/List_of_HTTP_header_fields
+ *
+ * @author Soufiane Ghzal <sghzal@gmail.com>
+ */
+class HeaderAppendPlugin implements Plugin
+{
+    private $headers = [];
+
+    /**
+     * @param array $headers headers to add to the request
+     */
+    public function __construct(array $headers)
+    {
+        $this->headers = $headers;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
+    {
+        foreach ($this->headers as $header => $headerValue) {
+            $request = $request->withAddedHeader($header, $headerValue);
+        }
+
+        return $next($request);
+    }
+}
--- a/src/HeaderDefaultsPlugin.php
+++ b/src/HeaderDefaultsPlugin.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Http\Client\Plugin;
+
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Set default values for the request headers.
+ * If a given header already exists the value wont be replaced and the request wont be changed.
+ *
+ * @author Soufiane Ghzal <sghzal@gmail.com>
+ */
+class HeaderDefaultsPlugin implements Plugin
+{
+    private $headers = [];
+
+    /**
+     * @param array $headers headers to set to the request
+     */
+    public function __construct(array $headers)
+    {
+        $this->headers = $headers;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
+    {
+        foreach ($this->headers as $header => $headerValue) {
+            if (!$request->hasHeader($header)) {
+                $request = $request->withHeader($header, $headerValue);
+            }
+        }
+
+        return $next($request);
+    }
+}
--- a/src/HeaderSetPlugin.php
+++ b/src/HeaderSetPlugin.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Http\Client\Plugin;
+
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Set headers to the request.
+ * If the header does not exist it wil be set, if the header already exists it will be replaced.
+ *
+ * @author Soufiane Ghzal <sghzal@gmail.com>
+ */
+class HeaderSetPlugin implements Plugin
+{
+    private $headers = [];
+
+    /**
+     * @param array $headers headers to set to the request
+     */
+    public function __construct(array $headers)
+    {
+        $this->headers = $headers;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
+    {
+        foreach ($this->headers as $header => $headerValue) {
+            $request = $request->withHeader($header, $headerValue);
+        }
+
+        return $next($request);
+    }
+}
--- a/spec/HeaderAppendPluginSpec.php
+++ b/spec/HeaderAppendPluginSpec.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace spec\Http\Client\Plugin;
+
+use PhpSpec\Exception\Example\SkippingException;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\StreamInterface;
+
+class HeaderAppendPluginSpec extends ObjectBehavior
+{
+    public function it_is_initializable()
+    {
+        $this->beConstructedWith([]);
+        $this->shouldHaveType('Http\Client\Plugin\HeaderAppendPlugin');
+    }
+
+    public function it_is_a_plugin()
+    {
+        $this->beConstructedWith([]);
+        $this->shouldImplement('Http\Client\Plugin\Plugin');
+    }
+
+    public function it_appends_the_header(RequestInterface $request)
+    {
+        $this->beConstructedWith([
+            'foo'=>'bar',
+            'baz'=>'qux'
+        ]);
+
+        $request->withAddedHeader('foo', 'bar')->shouldBeCalled()->willReturn($request);
+        $request->withAddedHeader('baz', 'qux')->shouldBeCalled()->willReturn($request);
+
+        $this->handleRequest($request, function () {}, function () {});
+    }
+}
--- a/spec/HeaderDefaultsPluginSpec.php
+++ b/spec/HeaderDefaultsPluginSpec.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace spec\Http\Client\Plugin;
+
+use PhpSpec\Exception\Example\SkippingException;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\StreamInterface;
+
+class HeaderDefaultsPluginSpec extends ObjectBehavior
+{
+    public function it_is_initializable()
+    {
+        $this->beConstructedWith([]);
+        $this->shouldHaveType('Http\Client\Plugin\HeaderDefaultsPlugin');
+    }
+
+    public function it_is_a_plugin()
+    {
+        $this->beConstructedWith([]);
+        $this->shouldImplement('Http\Client\Plugin\Plugin');
+    }
+
+    public function it_sets_the_default_header(RequestInterface $request)
+    {
+        $this->beConstructedWith([
+            'foo' => 'bar',
+            'baz' => 'qux'
+        ]);
+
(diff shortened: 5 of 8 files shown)


src/HeaderPlugin.php
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * Automatically removes the given header from the request.
+     *
+     * @param string $name name of the header to remove
+     */
+    public function withoutHeader($name)
+    {
+        $this->removedHeader[] = $name;
```
sagikazarmark: What do you think about replacing this property with using the headers property set to false? IMO that's not a valid value anyway.

src/HeaderPlugin.php
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Sets a header with the given value to every requests.
+     *
+     * @param string $name     name of the header
+     * @param string $value    value of the header
+     * @param bool   $override By default if the header is already set on the request the value wont be replaced
+     *                         pass $override to true to make the header value the same for every request
+     */
+    public function withHeader($name, $value, $override = false)
```
sagikazarmark: I am not sure about this method:
1. with prefix suggests that it is immutable, while it is not.
2. I see the point in a builder function, but plugins should be configured once and never touched again. Although requireing an array of header removals is probably harder (or not from configuration?),

/cc @joelwurtz

sagikazarmark:
@gsouf Thank you very much for your contribution. Sorry for the late reply, I was on holiday. I added some notes to the PR, can you answer them please? Thanks.

src/HeaderPlugin.php
gsouf: I called it withHeader and withoutHeader to make it look like the psr7 message interface. 

Do you prefer everything to be made at the instantiation?

src/HeaderPlugin.php
gsouf: I'm not sure to understand.
Do you want the header plugin to act in a single way? I mean either removing or adding header, and not both?

dbu:
hi, thanks for the pull request! i would prefer to inject arrays into the constructor over the methods. there is no need for the methods to change the plugin.

something like __construct($addHeaders, $overrideHeaders) with 2 hashmaps? add would be add if not there, override would not check if a header is already there, and if the value in override is `false`, the header is removed. having false as value in add would be an error, it makes no sense to do that.

wdyt?

sagikazarmark:
Do we need to validate input data? It will turn out during request send anyway, if it is invalid.

Also, what about header removal?

gsouf:
I dont think we need to validate the input data, imo that's  an unneeded overhead, and the developer should take care of that by himself. Additionally the messageInteface implementation may already validate it. The plugin is just the relay for the data!

gsouf:
About array in the constructor, It's the right way if we want it to be unmutable. 

Using `key=>value` as `Header-Name=>headerValue` is also a good way. 

However I think it can be confusing to have to pass 2 arrays.

An other approach would be to have in paramater 1 the array of headers and in parameter 2 if the override is allowed. However ff we want both of override and non-override header that will need to have 2 instances of the header plugin. e.g:

``` php

$pluginThatCanOverride = new HeaderPlugin(["foo"=>"bar"], true);
$pluginThatCannotOverride = new HeaderPlugin(["foo"=>"bar"], false);

```

Still an other approach that is more KISS would be to have 1 plugin class that overrides, 1 plugin class that does not override, and 1 plugin the removes. This way things are very distinct, but that does a lot of classes and could bring an other confusion. e.g:

``` php

$pluginThatCanOverride = new HeaderOverridePlugin(["foo"=>"bar"]);
$pluginThatCannotOverride = new HeaderNonOverridePlugin(["foo"=>"bar"]);
$pluginThatRemoves = new HeaderRemovePlugin(["foo"=>"bar"]);

```

Every approach would work, that's design decisions!

joelwurtz:
Maybe we can accept à boolean in the value of the header

if false remove header if present
if true check if header exists and throw an exception when header is not
here

so we can have a single array
Le 19 janv. 2016 1:27 PM, "Soufiane Ghzal" notifications@github.com a
écrit :

gsouf:
@joelwurtz can you provide an example please?

dbu:
i think two separate parameters for only add and for force overwrite are
the easiest to understand for the user. i prefer that over a nested
array. but we could also have just one parameter with arrays with each
the keys header, value, force (or just 3 elements that represent this,
to make things less verbose)

sagikazarmark:
I am even thinking about separating this plugin into three.

dbu:
or that, yes. likely people only need one of them to set one header. its
not like we usually add dozens of static headers all over the place.

sagikazarmark:
From a configuration point of view, it is probably easier:

``` yaml

httplug.header.add:
    Header: value
httplug.header.remove:
    - Header2
```

gsouf:
The 3 separated plugins looks good to me

sagikazarmark:
:+1:

gsouf:
@sagikazarmark  Do you want I implement it with 3 classes ?

sagikazarmark:
Yes please.

gsouf:
Please can you tell me if the name of the class are ok to you? 

I will fix styles once everything else is validated. Also, I would be happy if you tell me how I can automatically fix the styles?

src/HeaderAddPlugin.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Http\Client\Plugin;
+
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Adds headers to the request only if the request does not already own it.
```
dbu: s/own/have/

sagikazarmark:
LGTM, although it seems to be hard to make a difference between adding and setting.

gsouf:
@sagikazarmark yes, I dont have a good idea for the name, do you have some?

dbu:
i like it. this becomes very clear to read! i have no better idea than add and set, and the phpdoc is clear.

you can click on the styleci link in the checks at the bottom of this page. styleci should allow you to create a pull request on your pull request or download the raw diff that you can apply. once you connected styleci with your gitlab (the login button)

src/HeaderAddPlugin.php
gsouf: Should I replace the comment with the following: 

`Adds headers to the request only if the request hasn't it already.` ?

sagikazarmark: Is this actually true? What about appending, but not replacing headers?

dbu: it currently is true. i guess it depends on the header whether appending makes sense or not, but the code below only adds the header if its not there yet. for many, that does make sense.

src/HeaderSetPlugin.php
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
+    {
+        foreach ($this->headers as $header => $headerValue) {
+            $request = $request->withHeader($header, $headerValue);
```
dbu, line 31: @sagikazarmark will this in all cases replace the header? or could this also add to the header? is it an implementation detail? or should we do ->withoutHeader($header)->withHeader($header, $headerValue) to be sure?

src/HeaderAddPlugin.php
gsouf: I'm not sure about appending, I think it's an edge case, not sure, actually I have difficulties to imagine a real case.

src/HeaderSetPlugin.php
sagikazarmark, line 31: no, withHeader replaces any existing headers. Appending can be done using withAddedHeader.

src/HeaderAddPlugin.php
sagikazarmark: Well, I think add should always add. If exists, append, if not exists, add.

src/HeaderSetPlugin.php
gsouf, line 31: Yes it will replace, you can check here: http://www.php-fig.org/psr/psr-7/#headers-with-multiple-values and here http://www.php-fig.org/psr/psr-7/#case-insensitive-header-field-names

src/HeaderAddPlugin.php
gsouf: What should we do? A fourth class `HeaderAppendPlugin` ?

sagikazarmark: No, I think the behavior above can simply be implemented here. Just add an else to the if with the other behavior (or remove the negation and exchange them).

gsouf: Well, afterthough I think we should rename `HeaderAddPlugin` to `HeaderAppendPlugin`.

We can also implement `HeaderAddPlugin`, but I think it's an edge case, and maybe custom plugin should be used for that?

I'm based on the example of the user agent. By default I want an user agent string, but if the request already has an user agent string, I wont put 2 values for user agent, I will set the header only if the request doesn't already has one.

sagikazarmark: Hm, I see the problem. I would still keep this functionality here. I could imagine a HeaderDefaults plugin which only adds stuff if necessary. WDYT?

gsouf: Yes `HeaderDefaults` looks very clear!!

dbu:
good one, yes! so we can have 
- HeaderDefaultsPlugin: set header if not set
- HeaderAppendPlugin: append values to header if already set, create otherwise
- HeaderSetPlugin: set header, overwrite if existing
- HeaderRemovePlugin: remove if existing

dbu:
i would call it HeaderAppendPlugin over HeaderAddPlugin as "add" is ambigious in this context imho.

sagikazarmark:
:+1: 

Should these go under a Header namespace? Or just leave them as is?

dbu:
i would leave them here. they are still grouped together as the name
starts with Header, good enough for me.

gsouf:
Are we good with this structure?

gsouf:
Implemented, I'm just doubting about the headerAppend. 

For instance if I use the following:

`new HeaderAppendPlugin(['Accept' => 'application/json'])`

If the request already has a `Accept` header with `application/json` then it does not make sense to add it again. That's why I think that HeaderAppend mostly covers edge cases that most of time are better to be implemented by the user.

sagikazarmark:
Obviously there are cases where it doesn't make sense. We should document that. This rather make sense in case of headers, where multiple values are allowed.

src/HeaderDefaultsPlugin.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace Http\Client\Plugin;
+
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Set default values for the request headers.
+ * If a given header already exists the value wont be replaced
```
dbu: ... an the request is not changed.

dbu:
great, i think it is good like this. append is a special case yes, but there are some headers where this makes sense, e.g. X-Forwarded-For . we can handle that in documentation.

src/HeaderAppendPlugin.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Http\Client\Plugin;
+
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Adds headers to the request.
+ * If the header already exists the value will be appended to the current value
```
dbu: maybe add a line below this one saying "This only makes sense for headers that can have multiple values like 'Forwarded'". Forwarded seems the only standard request header that can have multiple values, at least as i read on https://en.wikipedia.org/wiki/List_of_HTTP_header_fields (and we could link this wikipedia article from the documentation as well)

dbu:
there are a few styleci things to fix as well. (really minor ones, but lets keep things green ;-) )

gsouf:
Ok things are going to be green! :)

Waiting for your final validation before I update the doc

dbu:
:+1:

sagikazarmark:
Not a requirement, but @gsouf you could add yourself as an author to docbloc, just for us to know who to blame...I mean for credit. :stuck_out_tongue:

dbu:
for the blame, there is also `git blame`. but the glory should be all
yours with the `@author` annotation ;-)

gsouf:
Now you can blame me :-)

sagikazarmark:
Yeah. :+1:

dbu:
thanks @gsouf !

gsouf:
You're welcome, no time for the documentation now, I will do it tomorow

PR #45, opened 2016-01-25
use options resolver where applicable, fix decoder plugin gzip handling

dbu:
follow up of the discussion in https://github.com/php-http/documentation/pull/76 - i propose to use the OptionsResolver for configuration options on plugins. This is both more readable on the calling side and more flexible when we need to add more options in the future.

i mixed in the same PR a fix for the deccoder plugin to not claim to support gzip if it does not support gzip. this might hide the problem of a missing gzip library a bit.

### DIFF
--- a/src/DecoderPlugin.php
+++ b/src/DecoderPlugin.php
@@ -9,9 +9,14 @@ use Http\Message\Encoding\InflateStream;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\StreamInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
  * Allow to decode response body with a chunk, deflate, compress or gzip encoding.
+ *
+ * If zlib is not installed, only chunked encoding can be handled.
+ *
+ * If Content-Encoding is not disabled, the plugin will add an Accept-Encoding header for the encoding methods it supports.
  *
  * @author Joel Wurtz <joel.wurtz@gmail.com>
  */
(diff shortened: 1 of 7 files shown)


sagikazarmark:
Huge :+1:

joelwurtz:
:+1:

src/DecoderPlugin.php
```diff
@@ -39,11 +52,13 @@ public function __construct($useContentEncoding = true)
      */
     public function handleRequest(RequestInterface $request, callable $next, callable $first)
     {
-        $request = $request->withHeader('TE', ['gzip', 'deflate', 'compress', 'chunked']);
+        $encodings = extension_loaded('zlib') ? ['gzip', 'deflate', 'compress'] : [];
 
-        if ($this->useContentEncoding) {
-            $request = $request->withHeader('Accept-Encoding', ['gzip', 'deflate', 'compress']);
+        if ($this->useContentEncoding && count($encodings)) {
```
dbu, line 57: is it correct that servers that respect Accept-Encoding will understand us when not sending this header? or should we send an empty header or something to say we can't handle encodings?

joelwurtz, line 57: Yes, from the RFC : 

> If the Accept-Encoding field-value is empty, then only the "identity" encoding is acceptable.

And "identity" encoding = no encoding

joelwurtz, line 57: Speak to quick

> If no Accept-Encoding field is present in a request, the server MAY assume that the client will accept any content coding. In this case, if "identity" is one of the available content-codings, then the server SHOULD use the "identity" content-coding, unless it has additional information that a different content-coding is meaningful to the client.

joelwurtz, line 57: So we should set a default header IMO and add identity to the encoding list.

PR #46, opened 2016-01-25
ask for identity encoding if gzip not available.

dbu:
see https://github.com/php-http/plugins/pull/45/files#r50675017

### DIFF
--- a/src/DecoderPlugin.php
+++ b/src/DecoderPlugin.php
@@ -52,9 +52,9 @@ class DecoderPlugin implements Plugin
      */
     public function handleRequest(RequestInterface $request, callable $next, callable $first)
     {
-        $encodings = extension_loaded('zlib') ? ['gzip', 'deflate', 'compress'] : [];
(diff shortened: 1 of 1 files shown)


src/DecoderPlugin.php
```diff
@@ -52,7 +52,7 @@ public function __construct(array $config = [])
      */
     public function handleRequest(RequestInterface $request, callable $next, callable $first)
     {
-        $encodings = extension_loaded('zlib') ? ['gzip', 'deflate', 'compress'] : [];
+        $encodings = extension_loaded('zlib') ? ['gzip', 'deflate', 'compress'] : ['identity'];
 
         if ($this->useContentEncoding && count($encodings)) {
```
joelwurtz: count should be removed, as it will always be true

dbu: ups, fixed.

joelwurtz:
:+1:

PR #53, opened 2016-01-27
Updated doc block to be coherent with the suggested PSR

### LINKED ISSUE #51: Use new doc block for documenting array values
See [this resource](https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#examples-12). Should we make doc blocks like this? 

``` php
    /**
     * @param UriInterface $baseUri
     * @param array        $config {
     *     @var bool $replace True will replace all hosts, false will only add host when none is specified.
     * }
     */
    public function __construct(UriInterface $baseUri, array $config = [])
    {
```

Nyholm:
This will fix https://github.com/php-http/plugins/issues/51

### DIFF
--- a/src/CachePlugin.php
+++ b/src/CachePlugin.php
@@ -32,13 +32,13 @@ class CachePlugin implements Plugin
     private $config;
 
     /**
-     * Available options for $config are
-     *  - respect_cache_headers: Whether to look at the cache directives or ignore them.
-     *  - default_ttl: If we do not respect cache headers or can't calculate a good ttl, use this value.
-     *
      * @param CacheItemPoolInterface $pool
      * @param StreamFactory          $streamFactory
-     * @param array                  $config
+     * @param array                  $config        {
+     *
+     *     @var bool $respect_cache_headers Whether to look at the cache directives or ignore them.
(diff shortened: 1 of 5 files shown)


sagikazarmark:
LGTM. What about Style CI then?

Nyholm:
StyleCI does not like this doc block. Should I comply with the patch that StyleCI provides? Or do you want to change the config in StyleCI?

sagikazarmark:
Good question, have no idea. We could turn off some docblock check, maybe that helps.

joelwurtz:
Certainly the align thing in docblock,

joelwurtz:
I added this in php style ci config : 

```
disabled:
    - phpdoc_align
    - phpdoc_separation
```

Then it was ok, bug is coming from separation but if we only remove this, then align has another bug :/

joelwurtz:
I think for the moment we should add the blank line proposed by the fixer, it's not a big problem, and we can try to do a PR on php-cs-fixer (at least an issue)

sagikazarmark:
Ok, let's not add those disable statements then. I agree we can live with that extra line.

dbu:
:+1:

Nyholm:
Updated

PR #65, opened 2016-03-04
Add request conditional plugin

sagikazarmark:
Use case: cache certain responses based on a request condition.

### DIFF
--- a/src/RequestConditionalPlugin.php
+++ b/src/RequestConditionalPlugin.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Http\Client\Plugin;
+
+use Http\Message\RequestMatcher;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Apply a delegated plugin based on a request match.
+ *
+ * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
+ */
+final class RequestConditionalPlugin implements Plugin
+{
+    /**
+     * @var RequestMatcher
+     */
+    private $requestMatcher;
+
+    /**
+     * @var Plugin
+     */
+    private $delegatedPlugin;
+
+    /**
+     * @param RequestMatcher $requestMatcher
+     * @param Plugin         $delegatedPlugin
+     */
+    public function __construct(RequestMatcher $requestMatcher, Plugin $delegatedPlugin)
+    {
+        $this->requestMatcher = $requestMatcher;
+        $this->delegatedPlugin = $delegatedPlugin;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function handleRequest(RequestInterface $request, callable $next, callable $first)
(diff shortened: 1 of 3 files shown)


src/RequestConditionalPlugin.php
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * @var Plugin
+     */
+    private $delegatedPlugin;
+
+    /**
+     * @param RequestMatcher $requestMatcher
+     * @param Plugin         $delegatedPlugin
+     */
+    public function __construct(RequestMatcher $requestMatcher, Plugin $delegatedPlugin)
```
Nyholm, line 29: I maybe want to use more than one plugin. I suggest allowing the second parameter to be an array of Plugins. Would that be possible?

sagikazarmark, line 29: I was also thinking about that. For that case I would add a separate
ChainPlugin.

Tobias Nyholm notifications@github.com ezt írta (2016. március 5.,
szombat):

Nyholm, line 29: Okey. Cool!

sagikazarmark:
Anyone review?

src/RequestConditionalPlugin.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+namespace Http\Client\Plugin;
+
+use Http\Message\RequestMatcher;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Apply a delegated plugin based on a request match.
```
dbu, line 9: Execute a plugin only if the RequestMatcher matches the request.

src/RequestConditionalPlugin.php
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
+ */
+final class RequestConditionalPlugin implements Plugin
```
dbu, line 13: i am not too happy with the name. the name does not really tell how the 3 terms go together. is this a plugin to request a conditional? or to make the request conditional?

but unfortunately i have no better name. ConditionalRequestPlugin certainly is worse.

dbu:
would this now need to go against client-commons, now that we move the plugins there?

src/RequestConditionalPlugin.php
sagikazarmark, line 13: Alternative could be RequestMatcherPlugin. It's kinda obvious, don't know why I didn't think about it earlier.

sagikazarmark:
Yes, it should. Wanted to finish it before closing and moving to common.

src/RequestConditionalPlugin.php
dbu, line 13: yep, RequestMatcherPlugin is a good name!

PR #68, opened 2016-03-18
Fix #67 Fixed Cookie creation in cookie plugin.

### LINKED ISSUE #67: CookiePlugin creates Cookie objects with a DateTime object instead of an max-age integer; causing an InvalidArgumentException
CookiePlugin::createCookie is incompatible with the latest Cookie object. 
https://github.com/php-http/message/blob/master/src/Cookie.php

Bouwdie:
fix #67

Fix applied by calculating the expireIn (max-age) seconds instead of expire datetime.
Expanded save spec by using a more complete cookie.

Use DATE_COOKIE constant instead of DateTime constant.
Expand the cookie header content with all parts.

### DIFF
--- a/spec/CookiePluginSpec.php
+++ b/spec/CookiePluginSpec.php
@@ -2,15 +2,15 @@
 
 namespace spec\Http\Client\Plugin;
 
-use Http\Promise\FulfilledPromise;
 use Http\Message\Cookie;
 use Http\Message\CookieJar;
+use Http\Promise\FulfilledPromise;
 use Http\Promise\Promise;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\UriInterface;
-use PhpSpec\ObjectBehavior;
-use Prophecy\Argument;
 
 class CookiePluginSpec extends ObjectBehavior
 {
@@ -146,7 +146,7 @@ class CookiePluginSpec extends ObjectBehavior
 
         $response->hasHeader('Set-Cookie')->willReturn(true);
         $response->getHeader('Set-Cookie')->willReturn([
-            'cookie=value',
+            'cookie=value; expires=Tuesday, 31-Mar-99 07:42:12 GMT; Max-Age=60; path=/; domain=test.com; secure; HttpOnly'
         ]);
 
         $request->getUri()->willReturn($uri);
@@ -156,5 +156,25 @@ class CookiePluginSpec extends ObjectBehavior
         $promise = $this->handleRequest($request, $next, function () {});
         $promise->shouldHaveType('Http\Promise\Promise');
         $promise->wait()->shouldReturnAnInstanceOf('Psr\Http\Message\ResponseInterface');
+    }
+
+    function it_throws_exception_on_invalid_expires_date(RequestInterface $request, ResponseInterface $response, UriInterface $uri)
+    {
+        $next = function () use ($response) {
+            return new FulfilledPromise($response->getWrappedObject());
+        };
+
+        $response->hasHeader('Set-Cookie')->willReturn(true);
+        $response->getHeader('Set-Cookie')->willReturn([
+            'cookie=value; expires=i-am-an-invalid-date;'
+        ]);
+
+        $request->getUri()->willReturn($uri);
+        $uri->getHost()->willReturn('test.com');
+        $uri->getPath()->willReturn('/');
+
+        $promise = $this->handleRequest($request, $next, function () {});
+        $promise->shouldReturnAnInstanceOf('Http\Promise\RejectedPromise');
+        $promise->shouldThrow('Http\Client\Exception\TransferException')->duringWait();
     }
 }
--- a/src/CookiePlugin.php
+++ b/src/CookiePlugin.php
@@ -2,6 +2,7 @@
 
 namespace Http\Client\Plugin;
 
+use Http\Client\Exception\TransferException;
 use Http\Message\Cookie;
 use Http\Message\CookieJar;
 use Psr\Http\Message\RequestInterface;
@@ -86,6 +87,8 @@ class CookiePlugin implements Plugin
      * @param $setCookie
      *
      * @return Cookie|null
+     *
+     * @throws \Http\Client\Exception\TransferException
      */
     private function createCookie(RequestInterface $request, $setCookie)
     {
@@ -97,7 +100,8 @@ class CookiePlugin implements Plugin
 
         list($name, $cookieValue) = $this->createValueKey(array_shift($parts));
 
-        $expires = 0;
+        $maxAge = null;
+        $expires = null;
         $domain = $request->getUri()->getHost();
         $path = $request->getUri()->getPath();
         $secure = false;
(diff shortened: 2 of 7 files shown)


Bouwdie:
I am sort a new to php-spec :smile: Therefor i did not know how to assert values in the cookie object, since the cookiejar can not be mocked (final). Any suggestions?

sagikazarmark:
Thanks for the report and the PR. I will review this shortly.

spec/CookiePluginSpec.php
```diff
@@ -146,7 +146,12 @@ function it_saves_cookie(RequestInterface $request, ResponseInterface $response,
 
         $response->hasHeader('Set-Cookie')->willReturn(true);
         $response->getHeader('Set-Cookie')->willReturn([
-            'cookie=value',
+            sprintf(
+                'cookie=value; expires=%s; Max-Age=60; path=/; domain=test.com; secure; HttpOnly',
+                (new \DateTime())->add(
+                    new \DateInterval('PT60S')
+                )->format(DATE_COOKIE)
```
sagikazarmark: Can you please use `\DateTime::COOKIE` instead?

src/CookiePlugin.php
```diff
@@ -109,11 +109,12 @@ private function createCookie(RequestInterface $request, $setCookie)
 
             switch (strtolower($key)) {
                 case 'expires':
-                    $expires = \DateTime::createFromFormat(DATE_COOKIE, $value);
+                    $expires = \DateTime::createFromFormat(DATE_COOKIE, $value)->getTimestamp()
```
sagikazarmark: I don't think the expires header should return an integer as well. It is defined as datetime.

Bouwdie: That would mean that the expires header sets, the actual expires header in the Cookie object (8th param). I do think this would be neat but its use is deprecated? It also would be a change in behaviour compared to the current master version.

sagikazarmark: > That would mean that the expires header sets, the actual expires header in the Cookie object (8th param).

If the expires header is actually set, I think it should be set in the Cookie object as well. Otherwise it is inconsistent. Whether it is deprecated or not, we should follow the cookie specifications. Also, the expires header itself is deprecated, not the expires property of the object.

> It also would be a change in behaviour compared to the current master version.

Why would it? The expires header is a datetime object per specification/documentation. Changing it to integer value seems to be a change in behavior.

Or am I missing something?

Bouwdie: Well, it looked like the original implementation tried to derive the expiration date from the max-age. But was actually passed as max-age param to the cookie object (original bug). My first commit tried to solve it the other way around, by deriving the max-age from the expiration header. My latest commit passes expiration and max-age to the cookie object as 8th and 3rd constructor arg.

spec/CookiePluginSpec.php
```diff
@@ -146,7 +146,12 @@ function it_saves_cookie(RequestInterface $request, ResponseInterface $response,
 
         $response->hasHeader('Set-Cookie')->willReturn(true);
         $response->getHeader('Set-Cookie')->willReturn([
-            'cookie=value',
+            sprintf(
+                'cookie=value; expires=%s; Max-Age=60; path=/; domain=test.com; secure; HttpOnly',
+                (new \DateTime())->add(
+                    new \DateInterval('PT60S')
```
dbu: is this in seconds? will that mean the test fails if between this and the execution, the internal clock moves to the next second? that would be a flaky test. maybe we should regex match to allow the next second as well (i guess its reasonable to assume that the test will complete within 1 second at max) something like (%s|%s) with  \DateInterval('PT60S') and  \DateInterval('PT61S')

dbu:
that indeed looks more robust to me. @sagikazarmark can you maybe help with phpspec / my fear of a test that will randomly fail?

spec/CookiePluginSpec.php
xabbuh: Symfony provides a way to safely test time related code in the PHPUnit bridge: http://symfony.com/doc/current/components/phpunit_bridge.html#time-sensitive-tests

Maybe this of some help.

sagikazarmark: Is this actually time critical? Or we just need any timestamp here?

Bouwdie: There are no additional checks on the expires DateTime value in the Cookie object. Therefore not time critical. We could put a fixed cookie date string.

dbu: that sure would be simpler. lets put a fixed date 10 years in the future
so it looks not weird with an already expired date :-D

sagikazarmark: Or PHP max integer timestamp. That's surely in the future.

Bouwdie: Added +100 years :smile: 
Creating a Datetime  object from max int  has some interesting behaviour anyways:
https://3v4l.org/OI9RN#v550
https://3v4l.org/eV5jI

src/CookiePlugin.php
```diff
@@ -109,11 +110,11 @@ private function createCookie(RequestInterface $request, $setCookie)
 
             switch (strtolower($key)) {
                 case 'expires':
-                    $expires = \DateTime::createFromFormat(DATE_COOKIE, $value);
+                    $expires = \DateTime::createFromFormat(\DateTime::COOKIE, $value);
```
dbu, line 116: looking at the test failure on hhvm, i wonder if hhvm can't handle dates that far in the future? 

which in turn leads me to think we should check the return value as [\DateTime::createFromFormat](http://php.net/manual/en/datetime.createfromformat.php) is one of those stupid functions that returns false instead of throwing an exception on failure. if we check for `false === $expires` we can throw an exception and put the $value in the message to say it was not converted to a date.

Bouwdie:
Hmm hhvm does not seem to handle 4 digit year in cookie expires date. 
https://3v4l.org/bmmjC

dbu:
oh well... hm, whatever then lets just put 99 :-D
not our problem to solve really, and in 2099 we probably won't live
anymore (and i would be surprised if PHP was still around then...). and
in 2100, 99 will probably mean 2199 so we are all good.

src/CookiePlugin.php
Bouwdie, line 116: And what exception type would you like this to be? It probably needs to implement the name namespace exception interface. Can it be a TransferException? Or since there is a response a HttpException?

sagikazarmark, line 116: I would say it is not an HTTP related exception, so we should just fail with one of the SPL exceptions. UnexpectedValue maybe?

dbu, line 116: if i am not mistaken, HttpException is to express a HTTP error code. so it would be TransferException and we say that the server sent unexpected data. 

looking at DateTime::COOKIE having "l, d-M-Y H:i:s T" and Y meaning 4 digit year, i have no idea what hhvm is doing here. but if it works for all versions with a two digit year lets go with that - its just mock data after all.

and lets add a test that triggers the exception, with a date value of "this is not a date" or something.

dbu, line 116: > I would say it is not an HTTP related exception, so we should just fail
> with one of the SPL exceptions. UnexpectedValue maybe?

well, its kinda inverse 400: the server sent an invalid response... so i
think TransferException sounds right.

sagikazarmark, line 116: Ok, fine from my side.

sagikazarmark:
One last minor thing (Style CI) and ready to merge from my point of view.

dbu:
:+1: maybe also squash the commits.

src/CookiePlugin.php
```diff
@@ -86,6 +87,7 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
      * @param $setCookie
      *
      * @return Cookie|null
```
dbu, line 89: the styleci complaint is here: we require a blank line between the `@return` and an `@throws`

Bouwdie:
Fixed and squashed :8ball:

sagikazarmark:
Merged this manually because there were some conflicts.

sagikazarmark:
Thanks @Bouwdie

PR #70, opened 2016-05-04
Move plugins

sagikazarmark:
Although there will be no version two, I added it so that in any case, we can continue using this package if necessary (like a metapackage installing all plugins, discussed elsewhere).

Also, please let me know what you think about the compatibility layer:
- Old plugin interface extends the new one, no API change
- Old plugin client wraps the new one

### DIFF
--- a/src/PluginClient.php
+++ b/src/PluginClient.php
@@ -2,43 +2,26 @@
 
 namespace Http\Client\Plugin;
 
-use Http\Client\Common\EmulatedHttpAsyncClient;
-use Http\Client\Exception;
+@trigger_error('The '.__NAMESPACE__.'\PluginClient class is deprecated since version 1.1 and will be removed in 2.0. Use Http\Client\Common\PluginClient instead.', E_USER_DEPRECATED);
+
 use Http\Client\HttpAsyncClient;
 use Http\Client\HttpClient;
 use Http\Client\Plugin\Exception\LoopException;
-use Http\Promise\FulfilledPromise;
-use Http\Promise\RejectedPromise;
 use Psr\Http\Message\RequestInterface;
-use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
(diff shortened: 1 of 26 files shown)


src/PluginClient.php
```diff
@@ -2,43 +2,24 @@
...
-use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
  * The client managing plugins and providing a decorator around HTTP Clients.
  *
  * @author Joel Wurtz <joel.wurtz@gmail.com>
+ *
+ * @deprecated since since version 1.1, and will be removed in 2.0. Use {@link \Http\Client\Common\PluginClient} instead.
  */
 final class PluginClient implements HttpClient, HttpAsyncClient
```
dbu, line 19: could we simply extend the base plugin client and add nothing? as the plugin interface extends the base plugin interface, everything should "just work". or is the base client final?

dbu:
looks good. and we actually should release that as 1.1 at some point. otherwise people that use stable deps will never see the deprecations in their code.

we could also add the trigger_error code like symfony does, to see deprecation errors when enabled. https://knpuniversity.com/blog/upgrading-symfony-2.7

src/PluginClient.php
sagikazarmark, line 19: No we can't:
- The base client is final
- We would still have to do some hacks to throw the correct exception (in which case the exception is also final, so extension is not an option there)

sagikazarmark:
> we could also add the trigger_error code like symfony does

Ah, wanted to, but forgot it.

sagikazarmark:
Merge and tag?

dbu:
:+1: i let you tag if all is ready.

sagikazarmark:
Hm, what else to do?

Only the separate packages are not tagged, everything else is.

dbu:
i think we should tag those (if they are ready to be tagged) before we tag this. when people discover they should move, the move target should be released and ready :-)

